PR #12, opened 2016-04-09
Add _y_spt_afterframes_await_load

mikesmiffy128:
Well, here's a thing. No idea how well it works and it probably can't time stuff accurately but at least you can execute a full TAS continuously without user input :laughing:

NOTE: Totally alpha, unresearched and with NO guarantee of accuracy. Better put that full disclaimer there so that I'm definitely not claiming that this is even useful for most things.

That said, it would be nice if you'd merge :stuck_out_tongue:
Though actually if you choose not to, I wouldn't mind just being able to have it as my own fork if you push the latest changes (because this branch isn't up to date... is it?)

### DIFF
--- a/spt/OrangeBox/modules/ClientDLL.cpp
+++ b/spt/OrangeBox/modules/ClientDLL.cpp
@@ -359,7 +359,8 @@ void ClientDLL::ResetAfterframesQueue()
 
 void ClientDLL::OnFrame()
 {
-	if (afterframesPaused) {
+	if (afterframesPaused)
+	{
 		return;
 	}
 
--- a/spt/OrangeBox/modules/ClientDLL.cpp
+++ b/spt/OrangeBox/modules/ClientDLL.cpp
@@ -335,6 +335,8 @@ void ClientDLL::Clear()
 	pCmd = 0;
 
 	afterframesQueue.clear();
+	afterframesPaused = false;
+
 	duckspam = false;
 
 	setPitch.set = false;
@@ -357,6 +359,10 @@ void ClientDLL::ResetAfterframesQueue()
 
 void ClientDLL::OnFrame()
 {
+	if (afterframesPaused) {
+		return;
+	}
+
 	for (auto it = afterframesQueue.begin(); it != afterframesQueue.end(); )
 	{
 		it->framesLeft--;
--- a/spt/OrangeBox/modules/ClientDLL.hpp
+++ b/spt/OrangeBox/modules/ClientDLL.hpp
@@ -59,7 +59,6 @@ public:
 
 	void PauseAfterframesQueue() { afterframesPaused = true; }
 	void ResumeAfterframesQueue() { afterframesPaused = false; }
-	//bool IsAfterframesQueuePaused() { return afterframesPaused; }
 
 	void EnableDuckspam()  { duckspam = true; }
 	void DisableDuckspam() { duckspam = false; }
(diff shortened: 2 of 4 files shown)


spt/OrangeBox/modules/EngineDLL.cpp
```diff
@@ -295,7 +295,7 @@ bool __cdecl EngineDLL::HOOKED_SV_ActivateServer_Func()
 			shouldPreventNextUnpause = true;
 		}
 	}
-
+
```
YaLTeR: Remove the extra spaces

spt/OrangeBox/modules/ClientDLL.hpp
```diff
@@ -57,6 +57,10 @@ class ClientDLL : public IHookableNameFilter
 	void AddIntoAfterframesQueue(const afterframes_entry_t& entry);
 	void ResetAfterframesQueue();
 
+	void PauseAfterframesQueue() { afterframesPaused = true; }
+	void ResumeAfterframesQueue() { afterframesPaused = false; }
+	//bool IsAfterframesQueuePaused() { return afterframesPaused; }
```
YaLTeR: Either uncomment or remove, if it's gonna be needed later it can be added.

spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -357,6 +359,10 @@ void ClientDLL::ResetAfterframesQueue()
 
 void ClientDLL::OnFrame()
 {
+	if (afterframesPaused) {
```
YaLTeR: The code around there puts the brace on a new line, do the same please

YaLTeR:
All changes that I made since I pushed this branch are lost on the dead HDD, unfortunately.

mikesmiffy128:
Fixed  most things except the weird trailing tab in in EngineDLL. Oddly my Git doesn't seem to show that. Maybe it's editable within GitHub; I don't really know.

Dead HDD sounds pretty bad. :frowning:
Is there any way to recover it? I have an old drive with a broken partition table and I got stuff off it with TestDisk. I don't know how bad a state your drive is in though.

Alternatively, how hard would it be to recreate all the changes? The main difference I can remember is pitch not working while strafing. Can't be *that* hard, right?

YaLTeR:
Yea wouldn't be that hard to recreate. The HDD is at the doesn't mount at all point.

mikesmiffy128:
Well, you could try TestDisk anyway; it's pretty powerful. I think it tries to read the disk directly so it even if most of it is totally wrecked you can probably get at least something back. Unless of course the actual hardware interface isn't even sending data.

Anyway, anything else I should add to this or can it be merged? :smiley:

YaLTeR:
Well, once again, please test everything (if you haven't).

mikesmiffy128:
Well, to the best of my knowledge it shouldn't fail horribly. It is pretty simple, just postpones afterframes. I still make no guarantee that it is totally reliable / accurate, but it shouldn't crash or anything like that. I can launch the game and play it and it works just fine (fine or course being relative to the Source Engine in general).

YaLTeR:
I'm more curious about the consistency because if it's inconsistent then there isn't much point in this.

mikesmiffy128:
Hmm... I don't really know how I'd test that. Well, I'll keep this open for now and once I figure out some kind of reproducible thing it I'll let you know.

YaLTeR:
Just make a tas script that goes through a changelevel / load and see if it ends up in the same spot every time you launch it

mikesmiffy128:
Well, I came up with a basic test; it just starts strafing, saves and loads and then does _y_spt_getvel, _y_spt_getangles and getpos.

Seems pretty consistent; I attached both the test script and 15 results I got (all the same as far as I can see) as well as the starting save I used. 

You might want to try it out in case it varies from system to system but it seems okay to me.

[multisegtest.zip](https://github.com/YaLTeR/SourcePauseTool/files/211845/multisegtest.zip)

PR #13, opened 2016-06-06
Add _y_spt_afterframes_await_activate

mikesmiffy128:
Probably not _that_ useful, but I thought it was worth having. Also fixes what looks to me like an issue where afterframes_await_\* would break demo_pause_on_tick.

Also, if another character shows up, I'm truly sorry. I tried getting rid of one but Git just doesn't like to deal with trailing whitespace apparently. I have a feeling if I check it out with my current settings it'll just remove itself anyway. You can do some mad hacks to remove it if you want, I'm not gonna try. :stuck_out_tongue:

### DIFF
--- a/spt/OrangeBox/modules/ClientDLL.hpp
+++ b/spt/OrangeBox/modules/ClientDLL.hpp
@@ -57,8 +57,11 @@ public:
 	void AddIntoAfterframesQueue(const afterframes_entry_t& entry);
 	void ResetAfterframesQueue();
 
-	void PauseAfterframesQueue() { afterframesPaused = true; }
-	void ResumeAfterframesQueue() { afterframesPaused = false; }
+	// The pause state is decremented at each loading stage. It must reach 0 for the afterframes queue to resume.
+	// Therefore, we set the state to 1 to unpause on ActivateServer, and 2 to unpause on FinishRestore
+	// (as FinishRestore comes last so requires the highest starting value).
+	void SetAfterframesPauseState(int state) { afterframesPauseState = state; }
+	void DecrementAfterframesPauseState() { if (afterframesPauseState > 0) afterframesPauseState--; }
 
 	void EnableDuckspam()  { duckspam = true; }
 	void DisableDuckspam() { duckspam = false; }
@@ -95,7 +98,7 @@ protected:
 	bool tasAddressesWereFound;
 
 	std::vector<afterframes_entry_t> afterframesQueue;
-	bool afterframesPaused = false;
+	unsigned int afterframesPauseState = 0;
 
 	bool duckspam;
 	angset_command_t setPitch, setYaw;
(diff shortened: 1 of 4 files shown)


spt/OrangeBox/modules/ClientDLL.hpp
```diff
@@ -57,8 +57,12 @@ class ClientDLL : public IHookableNameFilter
 	void AddIntoAfterframesQueue(const afterframes_entry_t& entry);
 	void ResetAfterframesQueue();
 
-	void PauseAfterframesQueue() { afterframesPaused = true; }
-	void ResumeAfterframesQueue() { afterframesPaused = false; }
+	// Pause afterframes queue by setting pause state...
```
YaLTeR: the afterframes queue, the pause state

spt/OrangeBox/modules/ClientDLL.hpp
```diff
@@ -57,8 +57,12 @@ class ClientDLL : public IHookableNameFilter
...
+	// Set state to 1 to unpause on ActivateServer, 2 to unpause on FinishRestore.
+	// Both of those functions decrement state; once it hits 0 it's unpaused.
+	// It's possible to add additional intermediate states by setting higher values, as long as it decrements to 0.
```
YaLTeR: add additional
add more perhaps

spt/OrangeBox/spt-serverplugin.cpp
```diff
@@ -1,3 +1,5 @@
+#include "stdafx.hpp"
```
YaLTeR: hm?

spt/OrangeBox/modules/ClientDLL.hpp
```diff
@@ -95,7 +99,7 @@ class ClientDLL : public IHookableNameFilter
 	bool tasAddressesWereFound;
 
 	std::vector<afterframes_entry_t> afterframesQueue;
-	bool afterframesPaused = false;
+	unsigned int afterframesPauseState = 0;
```
YaLTeR: Make sure this gets reset in Clear()

YaLTeR:
Are you sure any given load (loading a save; changelevel; third extreme case - loading a new map) will behave as it should?

mikesmiffy128:
Oh, don't know why I left that `stdafx.hpp` there. I think I did something to my project file since that file wasn't using a precompiled header. Any particular reason why? (I'll remove the include anyway).

The Clear thing is a good point. I'll fix those things in a moment. Haven't really extensively tested it to be honest. I'll probably need to do that tomorrow.

Guess the comments are a bit weirdly worded. And actually, looking at what I've written I realise it's probably too wordy as well. I'll try to make it more concise and also throw in some correct sentence structure.

mikesmiffy128:
Believe I've fixed stuff now, let me know if I've missed something. I'll hopefully do some testing tomorrow if I can be bothered.

mikesmiffy128:
Oh, I forgot to close this months ago, but it probably won't work so I'll just close it.

PR #20, opened 2017-05-23
Added y_spt_piwsave

Ramalayha:
Works like stucksave but saves when the prop loses collision with the wall.

### DIFF
--- a/spt/OrangeBox/modules/ClientDLL.cpp
+++ b/spt/OrangeBox/modules/ClientDLL.cpp
@@ -10,6 +10,12 @@
 #include <SPTLib\hooks.hpp>
 #include "ClientDLL.hpp"
 
+#if !defined(OE)
+using std::max;
+#define GAME_DLL
+#include "cbase.h"
+#endif
+
 using std::uintptr_t;
 using std::size_t;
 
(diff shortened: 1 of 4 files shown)


spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -415,6 +421,36 @@ void ClientDLL::ResetAfterframesQueue()
...
+	if (y_spt_piwsave.GetString()[0] != '\0')
+	{
+		int count = GetEntityCount();
+		for (int i = 0; i < count; ++i)
+		{
+			auto ent = GetEntity(i);
+			if (!ent)
+				continue;
+
+			auto phys = ent->VPhysicsGetObject();
```
YaLTeR: It’s not a very good idea to use these functions directly because their addresses change for pretty much every different game version, and the worst part is that there’s no failsafe: if you load this build of SPT on an incompatible game version and try using the CVar the game will either just crash or do something completely unpredictable.

spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -415,6 +421,36 @@ void ClientDLL::ResetAfterframesQueue()
...
 void ClientDLL::OnFrame()
 {
+#if !defined(OE)
+	extern int GetEntityCount();
+	extern CBaseEntity* GetEntity(int);
+
+	if (y_spt_piwsave.GetString()[0] != '\0')
+	{
+		int count = GetEntityCount();
+		for (int i = 0; i < count; ++i)
```
YaLTeR: Are you sure going through all entities every frame is a good idea? Perhaps there’s some function that we can hook that is called when something gets stuck in the wall.

Ramalayha: There probably is something that can be hooked, but going through the entity list once each frame is not something that causes lag.

Traderain:
You could try to find the function which draws the VPHYSICS COLLISION ERROR label thing and check what calls it. That way you wouldn't need to enumerate over every object each frame.

PR #27, opened 2017-08-02
Added prop-in-wall save

Ramalayha:
Automatically saves when you push a prop into a wall.

### DIFF
--- a/spt/OrangeBox/modules/ClientDLL.cpp
+++ b/spt/OrangeBox/modules/ClientDLL.cpp
@@ -10,6 +10,12 @@
 #include <SPTLib\hooks.hpp>
 #include "ClientDLL.hpp"
 
+#if !defined(OE)
+using std::max;
+#define GAME_DLL
+#include "cbase.h"
+#endif
+
 using std::uintptr_t;
 using std::size_t;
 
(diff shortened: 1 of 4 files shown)


spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -415,6 +421,43 @@ void ClientDLL::ResetAfterframesQueue()
...
+	if (y_spt_piwsave.GetString()[0] != '\0')
+	{
+		auto ply = GetEntity(1);
+		if (ply)
+		{
+			auto pphys = ply->VPhysicsGetObject();
+			if (pphys && (pphys->GetGameFlags() & FVPHYSICS_PENETRATING) == 0)
+			{
+				int count = GetEntityCount();
+				for (int i = 0; i < count; ++i)
```
mikesmiffy128: I might well be wrong here, but if worldspawn is 0 and the player is 1, isn't it unnecessary to start this loop at index 0, and you could use 2 instead? Not that it'd make a big difference anyway.

YaLTeR:
Alright so change of plans: since SPT's code base contains a bunch of stuff without checks that works only on one game version, this can go through as long as it doesn't crash the game when the cvar is switched off. Could you confirm that's the case? (and resolve the conflict?)

Margen67:
@Ramalayha :^)

YaLTeR:
Jukspa updated this over in #136

PR #41, opened 2018-01-18
Added SE 2013 TAS support and basic missinginfo support

TheProJamer:
All singleplayer games supported except Half-Life Source, which seemed to have rather extreme auto-strafe problems. I'm leaving that until I work out the issue, or someone who can help me does help me.
Added TAS support to:
- Half-Life 2 (build 2257546)
- Half-Life 2 Episode One and Two (build 2257546)
- Half-Life 2 Lost Coast (build 2257546)
- Portal (build 1910503)
Added basic SPT support to:
- Missing Information 1.6 (build 4104)
- Missing Information 1.4.7 (build 4044)


spt/OrangeBox/patterns.hpp
```diff
@@ -355,22 +363,6 @@ namespace Patterns
 			"xxxxxxxxxxxx????xxxxxxxxxxxxxxxxxx"
 		},
 
-		{
```
YaLTeR: Why was this removed?

TheProJamer: I have no idea.

YaLTeR: Add it back?

TheProJamer: After further examination, I've come to the conclusion that I removed it because the same two patterns are directly above it.

YaLTeR: Fair enough.

PR #50, opened 2018-08-01
Added vectorial strafing and y_spt_canjb

Jukspa:


spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -725,9 +758,7 @@ void __fastcall ClientDLL::HOOKED_AdjustAngles_Func(void* thisptr, int edx, floa
 		curState.LgagstFullMaxspeed = tas_strafe_lgagst_fullmaxspeed.GetBool();
 
 		auto pl = PlayerData();
-		CalcAbsoluteVelocity(player, 0);
-		float *vel = reinterpret_cast<float*>(reinterpret_cast<uintptr_t>(player) + offAbsVelocity);
-		pl.Velocity = Vector(vel[0], vel[1], vel[2]);
+		pl.Velocity = GetPlayerVelocity();
```
YaLTeR: Good refactoring :+1:

spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -773,7 +804,7 @@ void __fastcall ClientDLL::HOOKED_AdjustAngles_Func(void* thisptr, int edx, floa
 		{
 			bool cantjump = false;
 			// This will report air on the first frame of unducking and report ground on the last one.
-			if ((*reinterpret_cast<bool*>(reinterpret_cast<uintptr_t>(player) + offDucking)) && ((*reinterpret_cast<int*>(reinterpret_cast<uintptr_t>(player) + offFlags)) & FL_DUCKING)) {
+			if ((*reinterpret_cast<bool*>(reinterpret_cast<uintptr_t>(player) + offDucking)) && GetFlagsDucking()) {
```
YaLTeR: Here as well, thanks!

spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -806,26 +837,16 @@ void __fastcall ClientDLL::HOOKED_AdjustAngles_Func(void* thisptr, int edx, floa
 		}
 
 		Friction(pl, onground, vars);
-		Strafe(pl, vars, onground, jumped, ((*reinterpret_cast<int*>(reinterpret_cast<uintptr_t>(player) + offFlags)) & FL_DUCKING), type, dir, tas_strafe_yaw.GetFloat(), va[YAW], out, reduceWishspeed, btns, usingButtons);
 
-		//EngineDevMsg("[Strafing] Yaw = %.8f\n", out.Yaw);
+		if (tas_strafe_vectorial.GetInt() == 1)
```
YaLTeR: It's better to use `GetBool()` here since we only care about the cvar being "enabled" or "disabled".

spt/OrangeBox/modules/ClientDLL.hpp
```diff
@@ -18,6 +19,7 @@ typedef void(__fastcall *_CViewRender__OnRenderStart) (void* thisptr, int edx);
 typedef void*(__cdecl *_GetLocalPlayer) ();
 typedef void*(__fastcall *_GetGroundEntity) (void* thisptr, int edx);
 typedef void(__fastcall *_CalcAbsoluteVelocity) (void* thisptr, int edx);
+class Vector;
```
YaLTeR: Huh, is this really needed? Considering you're using it as a return value below, I'm not sure the compiler would accept an incomplete definition, in which case this line should be redundant. If it isn't, it's still better to include the corresponding header instead.

spt/strafestuff.cpp
```diff
@@ -256,38 +256,99 @@ double YawStrafeMaxAngle(PlayerData& player, const MovementVars& vars, bool ongr
...
+	if (tas_strafe_jumptype.GetInt() == 2) {
+		// OE bhop
+		out.Yaw = NormalizeDeg(tas_strafe_yaw.GetFloat());
+		out.Forward = false;
+		out.Back = false;
+		out.Right = false;
+		out.Left = false;
+		return true;
+	}
+	else if (player.Velocity.Length2D() >= vars.Maxspeed * ((ducking || (vars.Maxspeed == 320)) ? 0.1 : 0.5)) {
```
YaLTeR: It should be
```
} else if
```
and it got messed up due to default VS indentation settings.

spt/strafestuff.cpp
```diff
@@ -256,38 +256,99 @@ double YawStrafeMaxAngle(PlayerData& player, const MovementVars& vars, bool ongr
...
-				out.Yaw = NormalizeRad(atan2(vel.y, vel.x)) * M_RAD2DEG;
-				out.Forward = false;
-				out.Back = false;
-				out.Right = false;
-				out.Left = false;
-				return onground;
-			}
+			return true;
 		}
+		else if (tas_strafe_jumptype.GetInt() == 3) {
```
YaLTeR: `} else if`

spt/strafestuff.cpp
```diff
@@ -256,38 +256,99 @@ double YawStrafeMaxAngle(PlayerData& player, const MovementVars& vars, bool ongr
...
+	return false;
+}
+
+
+void StrafeVectorial(PlayerData& player, const MovementVars& vars, bool onground, bool jumped, bool ducking, StrafeType type, StrafeDir dir, double target_yaw, double vel_yaw, ProcessedFrame& out, bool reduceWishspeed)
+{
+	if (jumped && StrafeJump(player, vars, ducking, out)) {
+		MapSpeeds(out, vars);
+		return;
+	}
```
YaLTeR: Very nice refactoring!

spt/strafestuff.cpp
```diff
@@ -256,38 +256,99 @@ double YawStrafeMaxAngle(PlayerData& player, const MovementVars& vars, bool ongr
...
+		// Calculate updated yaw
+		double normalizedDiff = NormalizeDeg(target_yaw - vel_yaw + tas_strafe_vectorial_offset.GetFloat());
+		double additionAbs = std::min(static_cast<double>(tas_strafe_vectorial_increment.GetFloat()), std::abs(normalizedDiff));
+		double additionSign = (normalizedDiff > 0) ? 1 : -1;
+
+		// Snap to target if difference too large(likely due to an ABH)
+		if (std::abs(normalizedDiff) > tas_strafe_vectorial_snap.GetFloat() && tas_strafe_vectorial_increment.GetFloat() > 0)
+			out.Yaw = target_yaw;
+		else
+			out.Yaw = vel_yaw + additionAbs * additionSign;
```
YaLTeR: There's a hot [copysign](https://en.cppreference.com/w/cpp/numeric/math/copysign) function for this although I don't think it matters much.

spt/OrangeBox/spt-serverplugin.cpp
```diff
@@ -506,6 +506,46 @@ CON_COMMAND(y_spt_cvar2, "CVar manipulation, sets the CVar value to the rest of
...
+		{
+			Msg("Yes, projected landing height %.6f\n", player_origin.z);
+			return;
+		}
+
+		vel.z -= (gravity * frametime / 2);
+		player_origin.z += vel.z * frametime;
+		vel.z -= (gravity * frametime / 2);
+	}
+	Msg("No\n");
```
YaLTeR: Not sure I'm a fan of this "No" message; maybe something more extended like "Jumpbug is impossible"? The yes one can stay the same.

PR #64, opened 2019-04-04
noop character + hl2-tas tests

Jukspa:


spt/OrangeBox/scripts/framebulk_handler.cpp
```diff
@@ -207,6 +217,20 @@ namespace scripts
...
+			bool isNoop = true;
+
+			for (size_t i = 0; i < line.size(); ++i)
+			{
+				if (line[i] != NOOP)
+				{
+					isNoop = false;
+					break;
+				}
+			}
```
YaLTeR: So what this loop is doing is marking the section as noop if it's comprised completely of the `<` character. So even something like `<<<<<<-` wouldn't pass as noop because there's one non-`<` character (and who knows how it'd parse later). I think what you meant was the inverse: `bool isNoop = false;` and then `if (line[i] == NOOP) { isNoop = true; break; }`.

spt/OrangeBox/scripts/framebulk_handler.cpp
```diff
@@ -207,6 +217,29 @@ namespace scripts
...
+					allNotNoop = false;
+				}
+			}
+
+			if (isNoop)
+			{	
+				noopSections.insert(section);
+			}
+			else if(!allNotNoop)
+				throw std::exception("Only some characters were noop in block!");
```
YaLTeR:
```suggestion
				throw std::exception("Some characters in a noop block were not noop!");
```

Issue #68, opened 2019-07-08
portal 2 crash


rikupone:
with spt loaded in portal 2 and no cvars changed, jumping immediately crashes the game to desktop.

YaLTeR:
Oops, sounds like #67. cc @Jukspa pls unbreak

PR #69, opened 2019-07-19
Hotfix for portal 2 crash

Jukspa:
Fix for issue #68

Strings in 2007 source are just const char* pointers and presumably they changed that in portal 2 so the entity property code was crashing on trying to read strings.


YaLTeR:
Thanks!

spt/OrangeBox/modules/ClientDLL.cpp
```diff
@@ -658,16 +663,23 @@ Vector ClientDLL::GetPlayerEyePos()
...
 {
+	if (!ORIG_GetLocalPlayer)
+		return false;
 	return (*reinterpret_cast<int*>(reinterpret_cast<uintptr_t>(ORIG_GetLocalPlayer()) + offFlags)) & FL_DUCKING;
 }
 
 double ClientDLL::GetDuckJumpTime()
 {
+	if (!ORIG_GetLocalPlayer)
+		return 0;
```
Margen67: Why use both `0` and `false`?

Jukspa: Don't understand the question, if you can't access the local player data, then no valid value can be returned, in which case it returns 0/false. This is just a hotfix for crashes that are caused by calling a function pointer that is null.

YaLTeR: Huh? This function returns `double` so it's `0`, the other one returns `bool` so it's `false`.

Issue #15, opened 2017-01-21
Black Mesa


newuser543:
Hello!
I just recently learned about Source Pause Tool and was excited to test it out with Black Mesa.
I did place it in the bin folder and launched the game (with "-insecure"-parameter) as instructed.
Unfortunately, the game instantly closes as soon as i load the plugin. No error message is displayed in the game's console (like "unable to load spt" or so...).

I'm playing the Steam version of BM. I'm thankful for any help.

Margen67:
>Steam version of BM

Well there's the problem :^)

newuser543:
First of all, thanks for the reply!
So, there's no suitable version of the SPT for the Steam version (yet)? Is there a chance, that one will be made in the future?

YaLTeR:
Added spt-oe and spt-2013 dlls to the latest release. Try spt-2013.dll, but people are saying pretty much nothing works on Steam BM.

newuser543:
Thank you for the effort! Unfortunately - as you assumed - both spt-2013 and spt-oe don't work.
spt-2013 closes the game instantly, while spt-oe gives out the error message "Unable to load plugin...".

XcapeEst:
Do you happen to know why doesn't it work with the retail version? Now that we have a small community of runners, having it working would be great. Is it impossible?

YaLTeR:
Most likely because of the updated (and customized) engine requires new patterns or interfaces to be implemented. It's possible to make it work but someone needs to sit down and figure it out.

thisis2838:
Previous versions of Black Mesa works fine with all versions of SPT. [The Halloween Update](https://steamcommunity.com/app/362890/discussions/0/312265327167955178/) seems to have broken support for it and every single version since hasn't worked.

If anyone's looking into making this work for Black Mesa, then you'd want to start looking around in the Halloween Update. You can download it's binaries with 

_`download_depot 362890 362894 616421470133972767`_

PR #70, opened 2019-07-20
Latest BMS retail support

Jukspa:
Added BMS config
Added BMS jump pattern
Moved some cvars to prevent plugin load crash caused by too many convars in one file(yea lol idk either)

Fix to #15


spt/vgui/vgui_utils.hpp
```diff
@@ -6,9 +6,13 @@
 struct AudioState_t;
 #endif
 
-#include "inputsystem\buttoncode.h"
-
+#ifdef BMS
+class IClientMode;
```
YaLTeR: If it's not needed anywhere else maybe just leave `class IClientMode;` instead of `iclientmode.h` unconditionally? Just add a comment that it's needed for BMS.

Jukspa: ok fixed, looks like the interface isnt actually used directly in any of the gui code

YaLTeR:
Thanks, also the magic word is `fix #n` or `fixes #n`, resolve/s and close/s work too.

PR #76, opened 2019-07-27
Remove useless x64 config, change PlatformToolset to v141_xp

Margen67:
Source Engine is 32-bit anyway so x64 makes no sense.


YaLTeR:
Could you refresh the commit so appveyor picks it up? Not sure how to force a build.

YaLTeR:
Thanks!

PR #79, opened 2019-08-23
Set up CI with Azure Pipelines & AppVeyor fixups

Margen67:
Builds faster, and builds multiple configurations simultaneously.

Example:
https://dev.azure.com/Margen67/SourcePauseTool/_build

Also AppVeyor fixups :^)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,6 +3,7 @@ version: '{build}'
 skip_commits:
   files:
     - README.md
+    - .azure-pipelines.yml
     - .travis.yml
 
 os: Visual Studio 2017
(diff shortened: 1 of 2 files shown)


YaLTeR:
I've got no experience with Azure and as I said I don't see any value in testing the same thing on two different CI systems. It only adds maintenance overhead. So could you please pick one CI to leave?

Here are the options I'm aware of:
- ### Travis
  Travis recently added Windows support, and it's the one I have the most experience with. I'm not sure if they run Windows jobs in parallel, but if they do then it would probably be ideal since we already run the formatting in a job there.
- ### AppVeyor
  We already have this set up, and the benefit is that I have some experience with it, but obviously it's super slow.
- ### Azure Pipelines
  I don't have experience with this, but it looks like it gives 10 parallel jobs for free.
- ### CircleCI
  I also don't have experience with this but I know of some projects that used it.

PR #101, opened 2019-12-29
Saveloads in frame bulks and fix to afterframes resuming too late.

### LINKED ISSUE #91: [Request] Add support for recording demos through s/l
Currently there is no easy way to record demos through save-loads in a single script. It seems that the current recording mechanism is just using the record command, which stops as soon as you save-load. The only way to go recording demos like this atm is by using speedrun_segment which doesn't keep the same demo name as at the top of the script, or by doing some janky jankyness with y_spt_pause 2.

### LINKED ISSUE #96: _y_spt_afterframes_await_load resumes 1 tick after the game loads

Jukspa:
Fixes #96 and fixes #91.


YaLTeR:
Any tests for this?

Jukspa:
For the saveloads or afterframes resuming earlier?

YaLTeR:
Both, if possible, I guess?

Jukspa:
Gonna add a check so that the sl bulk doesnt overwrite the original save.

YaLTeR:
Thanks

PR #104, opened 2020-01-05
Improved tracing for ground detection

Jukspa:
Portal can now use normal tracing hulls instead of the line hack. Also fixed a bug with reducewishspeed not getting calculated correctly if player was ducking while < 2 units above the ground or if a player was near a portal while ducked.


YaLTeR:
Thanks!

Issue #72, opened 2019-07-24
"Error LNK1104 cannot open file 'libMinHook.x86.lib'" when cleanly building Debug


Margen67:
```Error LNK1104 cannot open file 'libMinHook.x86.lib'```
![image](https://user-images.githubusercontent.com/3462541/61777358-9f9a0a80-adb1-11e9-8082-738aed24eaf7.png)
If you build Release first, it builds successfully:
![image](https://user-images.githubusercontent.com/3462541/61777586-04edfb80-adb2-11e9-8d16-68bbc5110a76.png)

PR #110, opened 2020-02-08
Replace AppVeyor and Travis with GitHub Actions

Margen67:
### Why?
Because AppVeyor is slow.


YaLTeR:
Could you move the clang-format check from travis to GHA and remove travis and appveyor altogether? Also why are debug configs commented out?

Margen67:
Yes.
Debug configs fail to build from a clean build: #72

YaLTeR:
You can add the format job in the CI.yml file. Just put the whole format: section after build:. Also you can delete travis and appveyor configs altogether (and remove them from ignored paths in the yml).

Margen67:
@YaLTeR I did that and also uncommented Debug.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,84 @@
...
+            SDK: episode1
+            spt-filename: spt-oe
+          - Configuration: Debug P2
+            SDK: portal2
+            spt-filename: spt-p2
+    steps:
+    - name: Checkout AlliedModders SDK
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
YaLTeR: Why? I read that the default is 1, which should be sufficient?

Margen67: Does that apply here or in general?

YaLTeR: In general.

YaLTeR: (as long as it works)

YaLTeR:
And one last thing, could you check that the formatting job does indeed fail the build? Try committing some formatting error.

Margen67:
> 
> 
> And one last thing, could you check that the formatting job does indeed fail the build? Try committing some formatting error.

Like what?

YaLTeR:
Idk, like adding a bunch of spaces between arguments to some function.

Margen67:
Yup, formatting works: https://github.com/Margen67/SourcePauseTool/runs/433761083

YaLTeR:
Thanks for porting the CI!

YaLTeR:
hold up, it failed https://github.com/YaLTeR/SourcePauseTool/runs/434150058?check_suite_focus=true#step:6:10

Margen67:
It succeeds on my local repo so that makes no sense.
Besides reverting this PR, either this should be converted to use cmake or updated to VS2019.

PR #112, opened 2020-02-12
Changed the default value for y_spt_pause from 1 to 0

Jukspa:
Nowadays the plugin is used often for other purposes than pausing so it makes sense to switch this to 0. Other cvars for TASing and segmenting purposes default to off and y_spt_pause does not.


YaLTeR:
Yeah that's a good point.

PR #117, opened 2020-02-21
[CI] Add -j to submodule update

Margen67:
Saves ~1-5 seconds :^)


YaLTeR:
How? There's only one submodule, what does it do in parallel?

Margen67:
Might be placebo, then

YaLTeR:
I'll pass on this here then.

PR #119, opened 2020-02-22
Use project reference to take advantage of Link Library Dependencies

### LINKED ISSUE #72: "Error LNK1104 cannot open file 'libMinHook.x86.lib'" when cleanly building Debug

RedHaze:
Currently, the library `libMinHook.x86.lib` is always linked from the release build of libMinHook even when building SPT as Debug. This PR takes advantage of MSBuild [ProjectReference](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2019#projectreference) property and the `Link Library Dependencies` setting. By adding `libMinHook` as a project reference to `SPT`, the resulting library from `libMinHook` for the given configuration will be automatically linked.

Resolves #72

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,24 +14,23 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        Configuration: [Release, Release BMS, Release 2013, Release OE, Release P2]
-#        Configuration: [Debug, Debug BMS, Debug 2013, Debug OE, Debug P2, Release, Release BMS, Release 2013, Release OE, Release P2]
+        Configuration: [Debug, Debug BMS, Debug 2013, Debug OE, Debug P2, Release, Release BMS, Release 2013, Release OE, Release P2]
         include:
-#          - Configuration: Debug
-#            SDK: orangebox
-#            spt-filename: spt
-#          - Configuration: Debug BMS
-#            SDK: bms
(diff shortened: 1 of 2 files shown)


YaLTeR:
Oh, that's cool, thanks!

Could you squash the first 4 commits, and in the last one change the artifact name `name: ${{ matrix.spt-filename }}` to say "debug" if it's debug?

RedHaze:
> Oh, that's cool, thanks!
> 
> Could you squash the first 4 commits, and in the last one change the artifact name `name: ${{ matrix.spt-filename }}` to say "debug" if it's debug?

I'm assuming the second request is to prevent release & debug artifacts from clobbering each other? If that's the case, I made it so the artifact name contains the MSBuild configuration. I was going to do something like `if: contains(matrix.Configuration, 'debug')` but then I would need two copy artifact steps since (from the 30 minutes of research I've done) it doesn't seem like you can do inline conditionals. Otherwise, I would have done something like `name: ${{ matrix.spt-filename }} + contains(matrix.Configuration, 'debug') ?  '-debug' : ''`, however, I think including the configuration in the artifact name is more robust. Let me know what you think 😄 .

PR #124, opened 2020-04-04
Fixed first tick crashes with conditions

Jukspa:
New afterframes change broke some condition stuff resulting in typical null pointer shenanigans with non-existent entities. Also changed the serverActive function name to more correctly match its new implementation of just checking whether or not the player entity is available. That was what it was used for checking anyway, since all the functions using that function were reading some player property stuff.


YaLTeR:
Thanks, tell me if/when you need a release.

PR #125, opened 2020-04-10
Added micro-AFH

Jukspa:
Not sure if this should be handled as another jumptype instead


spt/strafestuff.cpp
```diff
@@ -41,6 +41,9 @@ ConVar tas_strafe_version("tas_strafe_version",
                           FCVAR_TAS_RESET,
                           "Strafe version. For backwards compatibility with old scripts.");
 
+ConVar tas_strafe_afh_length("tas_strafe_afh_length", "0.0000000000000000001", FCVAR_TAS_RESET, "Magnitude of AFHs");
```
YaLTeR, line 44: Is this something you can actually input? Also, is this cvar needed? Maybe just hardcode the value

spt/strafestuff.cpp
```diff
@@ -1028,7 +1042,7 @@ namespace Strafe
 	{
 		if (StrafeJump(jumped, player, vars, ducking, out))
 		{
-			if (!yawChanged || tas_strafe_allow_jump_override.GetBool())
+			if ((!yawChanged || tas_strafe_allow_jump_override.GetBool()) && !out.Processed)
```
YaLTeR, line 1045: Is this change related?

spt/strafestuff.cpp
Jukspa: yes and I figured I'd leave it in case there's a problem with the default value of the cvar not working for whatever reason in some scenario

spt/strafestuff.cpp
Jukspa: the other jump commands run the MapSpeeds function that sets the fmove and smove for that frame so that needs to be skipped with this method

Jukspa: Would you prefer if this got refactored into the StrafeJump function instead?

spt/strafestuff.cpp
Jukspa: I had some instances where it would work for some jumps and not for others and I'm not really sure why

YaLTeR: Oh, okay. I mean, it is pretty small, so there's a good chance that it underflows to 0 on some operation.

spt/strafestuff.cpp
YaLTeR: I see. Idk, whatever you think is better.

Jukspa: I think I'll refactor it

Jukspa:
done

PR #131, opened 2020-05-28
Make _y_spt_autojump_ensure_legit a cheat

Margen67:


YaLTeR:
Sure, thanks

Jukspa:
Marking a cvar with the flag FCVAR_CHEAT does nothing too interesting, it just marks it as something to reset when you sv_cheats 0, see for example r_drawclipbrushes. It doesn't actually prevent its use when cheats are enabled..

PR #138, opened 2020-06-12
Various changes

### LINKED ISSUE #137: y_spt_cvar_random uses the game's random stream
y_spt_cvar_random uses the game's random stream for generating cvar values which might alter the rng in the game. It might also be the cause of crashes in the clip script. Would probably be a good idea to make it use the standard library random number generators like the TAS search functionality does.

Jukspa:
Fixes #137 .


README.md
```diff
@@ -36,20 +36,25 @@ A plugin for all your pausing needs.
 ### Building
 You will need Visual Studio 2019 and [git](https://git-scm.com).
 
-1. Run the following in cmd:
+1. Open Visual Studio 2019(this was tested on the "Community" version but others probably work as well). Click on Tools -> Get Tools and Features... from the top bar of the window.
```
YaLTeR, line 39:
```suggestion
1. Open Visual Studio 2019. Click on Tools → Get Tools and Features... from the top bar of the window. 
```

Jukspa:
threw this HLS fix in here as well

Jukspa:
as a note, HLS is missing one member from CUserCmd which is causing the difference in the size, see: https://github.com/alliedmodders/hl2sdk/blob/0ef5d3d482157bc0bb3aafd37c08961373f87bfd/game/shared/usercmd.h#L173

YaLTeR:
Awesome, thanks!

Issue #92, opened 2019-11-10
y_spt_hud_oob crashes game during demo playback


UncraftedName:
Using the Aug 23 build

Jukspa:
yea this is unforunately the case. it would require a complete overhaul as to how it's calculated to fix this

Jukspa:
looks fixed as bill said, not entirely sure what did it but there you go

SirWillian:
The problem before was that GetPlayerEyePos would calculate the eye position using the server-side player, which causes problems when replaying demos, while now the camera position (that shows on cl_showpos) is being used

PR #139, opened 2020-06-12
Changed y_spt_hud_oob to use the camera position as the reference point

SirWillian:
This change adheres to the (mostly) agreed definition of Inbounds used by the Portal community. Doesn't seem like any other community is using this functionality to have much of an input on the matter


spt/OrangeBox/modules/ClientDLL.hpp
```diff
@@ -49,6 +49,7 @@ typedef void(__cdecl* _UTIL_TraceRay)(const Ray_t& ray,
                                       trace_t* ptr);
 typedef bool(__fastcall* _CGameMovement__CanUnDuckJump)(void* thisptr, int edx, trace_t& ptr);
 typedef void(__fastcall* _CViewEffects__Fade)(void* thisptr, int edx, void* data);
+typedef void*(__cdecl* _MainViewOrigin)();
```
Jukspa, line 52:
```suggestion
typedef const Vector&(__cdecl* _MainViewOrigin)();
```

SirWillian:
I noticed that this also fixed issue #92 . Didn't know it was an issue to begin with or else i'd have linked it sooner

PR #140, opened 2020-06-14
Fixed overlay camera and y_spt_hud_oob for 3420

SirWillian:
Added a couple patterns for those to work in Portal build 3420


SirWillian:
Also, I mentioned this a bit late on the previous PR, but issue #92 got fixed. Just restating here to make sure it's seen

PR #143, opened 2020-06-21
Added search condition for being in portal bubble

UncraftedName:
usage: `portal_bubble (1 or 0)`
Can use this for checking if you have or cancelled sg.

### DIFF
--- a/spt/OrangeBox/scripts/condition.cpp
+++ b/spt/OrangeBox/scripts/condition.cpp
@@ -10,7 +10,6 @@
 #include "..\modules.hpp"
 #include "..\modules\ClientDLL.hpp"
 
-
 namespace scripts
 {
 	TickRangeCondition::TickRangeCondition(int low, int high, bool reverse)
--- a/spt/OrangeBox/scripts/condition.cpp
+++ b/spt/OrangeBox/scripts/condition.cpp
@@ -8,6 +8,8 @@
 #include "..\modules.hpp"
 #include "..\modules\ClientDLL.hpp"
 
+const int INDEX_MASK = MAX_EDICTS - 1;
+
 namespace scripts
 {
 	TickRangeCondition::TickRangeCondition(int low, int high, bool reverse)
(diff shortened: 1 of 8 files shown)


Jukspa:
the plan in the long term would be generalizing the conditions so you can add logical operators and stuff like that, I was testing that in https://github.com/Jukspa/SourcePauseTool/tree/conditions

UncraftedName:
So is part of the goal to add search conditions for arbitrary entity properties?

Jukspa:
That's one option but the main problem with it is that you can't guarantee that the condition is valid on an arbitrary entity index and property since you can't know when parsing the script what the class of the entity is. Initially I figured I would just allow using known properties for the player and be able to combine them with logical operators. It's a lot more readable to write `if player.position[0] > 300 && player.position[1] >= 200` as opposed to using the posx and posy conditions.

spt/OrangeBox/scripts/srctas_reader.cpp
```diff
@@ -459,6 +459,9 @@ namespace scripts
 		propertyHandlers["changelevel"] = &SourceTASReader::HandleCLCondition;
 		propertyHandlers["velyaw"] = &SourceTASReader::HandleVelYaw;
 		propertyHandlers["velpitch"] = &SourceTASReader::HandleVelPitch;
+#if SSDK2007
+		propertyHandlers["portal_bubble"] = &SourceTASReader::HandlePBubbleCondition;
```
Jukspa, line 463:
```suggestion
		if (DoesGameLookLikePortal())
		{
			propertyHandlers["portal_bubble"] = &SourceTASReader::HandlePBubbleCondition;
		}
```
I'm not sure if spt-serverplugin.hpp is included but the function is declared there iirc.

spt/OrangeBox/scripts/condition.cpp
```diff
@@ -2,12 +2,16 @@
...
 
+#include <spt\OrangeBox\overlay\portal_camera.hpp>
+
 #include "..\..\utils\ent_utils.hpp"
 #include "..\..\utils\math.hpp"
 #include "..\..\utils\property_getter.hpp"
 #include "..\modules.hpp"
 #include "..\modules\ClientDLL.hpp"
 
+const int INDEX_MASK = MAX_EDICTS - 1;
```
Jukspa, line 13:
```suggestion
```
Looks like this is unused.

PR #151, opened 2020-09-08
Added basic IPC stuff

Jukspa:
Mostly a copy paste of the IPC stuff I used for TASQuake.

Using winsock, data goes in and out of the socket as null-terminated JSON strings. JSON-library used is https://github.com/nlohmann/json. At the moment you can only send console commands to the game and no way to send stuff back to the client but I figured I'd PR this minimal version first.

I had some issues with the clang format complaining about include order and suggesting one that would cause compilation errors with macros so I just removed it. It's pretty random anyway and half the files arent even ordered by the clang-format logic but even so it somehow doesn't find an issue with that until you change a filename and then it just suggests a completely different order.


YaLTeR:
Thanks!

PR #154, opened 2020-09-17
Some IPC fixes and commands

Jukspa:
I added the ACK message since I was having some issue in client programs terminating too early after sending a packet which wouldn't be transmitted. Previously I avoided this by adding an arbitrary wait after sending the last packet but that's probably not too smart. I think this is because TCP sends the ACK after you place the packet in the buffer but doesn't necessarily wait for the server to receive it or some bullshit like that, so I needed an application level ACK.

### DIFF
--- a/spt/OrangeBox/modules/inputSystemDLL.cpp
+++ b/spt/OrangeBox/modules/inputSystemDLL.cpp
@@ -1,0 +1,72 @@
+#include "stdafx.h"
+
+#include "inputsystemDLL.hpp"
+
+#include "convar.h"
+
+#include "..\modules.hpp"
+#include "..\patterns.hpp"
+
+#define DEF_FUTURE(name) auto f##name = FindAsync(ORIG_##name, patterns::inputsystem::##name);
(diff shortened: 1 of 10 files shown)


spt/OrangeBox/modules/inputSystemDLL.cpp
```diff
@@ -0,0 +1,72 @@
...
+				} \
+			} \
+		} \
+		else \
+		{ \
+			DevWarning("[inputsystem dll] Could not find " #future_name ".\n"); \
+		} \
+	}
+
+ConVar y_spt_focus_nosleep("y_spt_focus_nosleep", "0", 0, "Disables the FPS taking a shit while alt-tabbed.");
```
YaLTeR, line 35: more professional message please

YaLTeR (changes requested):

spt/OrangeBox/modules/inputSystemDLL.cpp
Jukspa: :(

YaLTeR:
Thanks

PR #156, opened 2020-11-23
tas_aim

Jukspa:


spt/utils/ent_utils.cpp
```diff
@@ -528,16 +528,7 @@ namespace utils
...
-	{
-		auto engine_server = GetEngine();
-		if (!engine_server)
-			return 0;
-
-		return engine_server->GetEntityCount();
-	}
-
+#ifndef P2
 	static CBaseEntity* GetServerEntity(int index)
```
YaLTeR, line 532: Is it intentional that this function is now compiled on OE?

Jukspa: It was intentional but OE is missing the necessary patterns for the tas_aim code anyway, so it hasn't been tested. I set it like that to futureproof it since there's nothing stopping it from compiling under OE, so I figure it would also work there.

YaLTeR:
Thanks!

PR #161, opened 2021-02-14
Build improvements

Margen67:


spt.vcxproj
```diff
@@ -255,18 +260,28 @@
...
+    git submodule update --init -j %NUMBER_OF_PROCESSORS% %SUBMODULE% &amp;&amp;^
+    echo Successfully checked out %SUBMODULE%.                                                       &amp;&amp;^
+    echo Please reload all projects.
+    EXIT /b 1
+)
+IF NOT EXIST SDK\orangebox\.git (
+    SET SUBMODULE=SDK\orangebox
+    git submodule update --init -j %NUMBER_OF_PROCESSORS% %SUBMODULE% &amp;&amp;^
+    echo Successfully checked out %SUBMODULE%.
+)</Command>
```
YaLTeR, line 274: I'm not sure I like checking out submodules from a pre-build event. I certainly wouldn't expect something like this in a pre-build event (although I don't deal with MSVC projects all that much, maybe it's actually common?). I'd rather leave it as-is.

spt.vcxproj
```diff
@@ -45,23 +45,26 @@
...
     <ProjectGuid>{B6572CBE-7C7F-4FFF-94DE-AFBBBAB11C64}</ProjectGuid>
-    <WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.19041.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <PlatformToolset>v141_xp</PlatformToolset>
     <CharacterSet>MultiByte</CharacterSet>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
```
YaLTeR, line 55: You should run the TAS tests (or get someone to do it), this kind of flag might influence the outcome (it'll probably be fine but still).

YaLTeR:
Thanks, I guess the format action needs to be updated, but that seems unrelated.

PR #162, opened 2021-03-14
Added _y_spt_afterframes for HL1 Movement, y_spt_autojump for Transmissions: Element 120

SmileyAG:


YaLTeR:
Just to be on the safe side I'd use more bytes for the new patterns, like in other mods. So that it definitely doesn't clash with something in other mod dlls.

YaLTeR:
Thanks

PR #165, opened 2021-03-16
Added autojump for ICE modification, changed pattern for y_spt_disabl…

SmileyAG:
…e_fade.

In the previous PR, I was very inattentive, and did not notice that the pattern from DMoMM for y_spt_disable_fade is also valid for the 4104/5135 version.

Now I am more attentive and I check **all** my actions before request the updates into official release, so I replaced it with the pattern from the Steampipe version. 

I also found pattern for wrong function in previous PR, but now it's done for right function. 

Excuse me.


YaLTeR:
Thanks

PR #166, opened 2021-03-18
Added y_spt_disable_shake

SmileyAG:
You may have a question: why is this cvar needed, can't you just use **shake_stop**?

**shake_stop** needs to be input manually with each new shake.

**y_spt_disable_shake** automatically disables all shake.

### DIFF
--- a/spt/OrangeBox/patterns.hpp
+++ b/spt/OrangeBox/patterns.hpp
@@ -268,6 +268,7 @@ namespace patterns
 		    "51 56 6A 14 8B F1 E8 ?? ?? ?? ?? 8B 54 24 10 8B C8 0F B7 02 89 44 24 10 83 C4 04 89 4C 24 04 DB 44 24 0C",
 		    "5377866",
 		    "55 8B EC 51 56 57 6A 14 89 4D FC E8 ?? ?? ?? ?? 8B 7D 08 8B F0");
+		PATTERNS(CViewEffects__Shake, "4104-5135", "56 8B 74 24 ?? 8B 06 85 C0");
 		PATTERNS(
 		    CHudDamageIndicator__GetDamagePosition,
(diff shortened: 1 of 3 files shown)


spt/OrangeBox/patterns.hpp
```diff
@@ -268,6 +268,7 @@ namespace patterns
 		    "51 56 6A 14 8B F1 E8 ?? ?? ?? ?? 8B 54 24 10 8B C8 0F B7 02 89 44 24 10 83 C4 04 89 4C 24 04 DB 44 24 0C",
 		    "5377866",
 		    "55 8B EC 51 56 57 6A 14 89 4D FC E8 ?? ?? ?? ?? 8B 7D 08 8B F0");
+		PATTERNS(CViewEffects__Shake, "4104-5135", "56 8B 74 24 ?? 8B 06 85 C0");
```
YaLTeR, line 271: Could you make the pattern a little longer pls? With the number of DLLs SPT has to run on I think it's best to be a bit more cautious about collisions.

YaLTeR (changes requested):

spt/OrangeBox/patterns.hpp
SmileyAG: Done.

SmileyAG:
Oops. Did a bit mistake in title of last commit, function is the same as before.

spt/OrangeBox/patterns.hpp
YaLTeR: Why did the `??` disappear?

SmileyAG: Well, I just copied the pattern and it works stably, I can back `??` if needed.

YaLTeR:
Thanks

PR #167, opened 2021-03-25
VAG crash fix

UncraftedName:
Added the `y_spt_prevent_vag_crash` var to prevent the "no free edicts" crash when trying a vertical angle glitch which makes searching for them slightly less painful. This also makes using a script for this kind of search (such as one through spt_ipc) much less annoying to use.

This kind of crash happens when a VAG teleports an entity slightly behind the entry portal after the 2nd teleport; which can cause a 4th teleport right back to where the VAG started. This would eventually cause a stack overflow, but since a physicsshadowclone gets created every time a teleport happens this way, the edicts all get used up without being freed for long enough.

To fix this I hooked the teleport function before and after the recursive call, and use a counter to keep track of the recursive depth. If the depth exceeds 3 (which is the max depth for a normal VAG), I nudge the teleported entity so that it ends up in front of the next portal it teleports to, breaking the cycle.


YaLTeR (approved):

YaLTeR:
Thanks!

PR #168, opened 2021-04-20
ISG toggle

UncraftedName:
- Added `y_spt_set_isg` for setting the ISG state
- Added `y_spt_hud_isg` for showing ISG state

Works for Portal 1 5135 ~~and 3420~~, as well as HL2 5135.

Also added a module for vphysics.dll, although I didn't end up using any hooking for this implementation.

### DIFF
--- a/spt/OrangeBox/modules/vphysicsDLL.cpp
+++ b/spt/OrangeBox/modules/vphysicsDLL.cpp
@@ -67,11 +67,15 @@ void VPhysicsDLL::Hook(const std::wstring& moduleName,
 	m_Length = moduleLength;
 	patternContainer.Init(moduleName);
 
-#ifdef SSDK2007
-	this->isgFlagPtr = (bool*)moduleBase + 0xe04d4;
-#else
-	Warning("y_spt_hud_isg 1 and y_spt_set_isg have no effect\n");
-#endif // SSDK2007
+	uint32_t ORIG_MiddleOfRecheck_ov_element = NULL;
+
+	DEF_FUTURE(MiddleOfRecheck_ov_element);
+	GET_FUTURE(MiddleOfRecheck_ov_element);
+
+	if (ORIG_MiddleOfRecheck_ov_element)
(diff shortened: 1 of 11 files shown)


spt/OrangeBox/modules/vphysicsDLL.cpp
```diff
@@ -0,0 +1,64 @@
...
+                       bool needToIntercept)
+{
+	Clear();
+	m_Name = moduleName;
+	m_Base = moduleBase;
+	m_Length = moduleLength;
+	patternContainer.Init(moduleName);
+
+#ifdef SSDK2007
+	this->isgFlagPtr = (bool*)moduleBase + 0xe04d4;
```
YaLTeR, line 48: No pattern, just a hardcoded offset? I guess if there's no hooking then it can only crash on read or write so not a very big deal

UncraftedName:
I just realized that the pattern I used works for 3420 but that is not the same as the hard-coded offset I implemented in these commits. I will add an offset for 3420 in the next couple of days.

UncraftedName:
Should be good to go now.

spt/OrangeBox/modules/vphysicsDLL.cpp
```diff
@@ -0,0 +1,83 @@
...
+	m_Base = moduleBase;
+	m_Length = moduleLength;
+	patternContainer.Init(moduleName);
+
+	// use patterns of hooked functions in clientDLL to figure out which version we're on and therefore which offset to use
+	int isgOff = 0;
+#if defined(SSDK2007)
+	if (clientDLL.ORIG_UTIL_TraceRay)
+	{
+		std::string vers = clientDLL.patternContainer.FindPatternName((PVOID*)&clientDLL.ORIG_UTIL_TraceRay);
```
YaLTeR, line 52: Kinda weird deciding vphysics.dll offsets based on client.dll, for example it means it might break for mods unnecessarily. But I'll leave the decision to @Jukspa.

UncraftedName:
I changed it to use a pattern now, and also amended the previous commits a bit as there were some mistakes.

YaLTeR:
Thanks!

PR #172, opened 2021-07-07
y_spt_disable_tone_map_reset

UncraftedName:
Useful for rendering runs with several demos, allows the color to transition smoothly from one demo to the next.

`y_spt_disable_tone_map_reset 0` (default behavior):

value at 0

`y_spt_disable_tone_map_reset 1`:

value at 1

### DIFF
--- a/spt/OrangeBox/cvars.cpp
+++ b/spt/OrangeBox/cvars.cpp
@@ -25,6 +25,11 @@ ConVar y_spt_on_slide_pause_for("y_spt_on_slide_pause_for",
                                 "0",
                                 0,
                                 "Whenever sliding occurs in DMoMM, pause for this many ticks.");
+ConVar y_spt_disable_tone_map_reset(
+    "y_spt_disable_tone_map_reset",
+    "0",
+    0,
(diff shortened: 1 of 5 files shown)


SmileyAG:
Could you add pattern for SDK2013-based games / mods? (Half-Life 2 Steampipe e.g.)

I already found that one, ``55 8B EC 8B 0D ?? ?? ?? ?? 56 8B 01 FF 90 88 01 00 00 8B F0 85 F6 74 07 8B 06 8B CE FF 50 08 8B 06 D9 45 08``, @UncraftedName

YaLTeR (approved):
Oh, this is cool.

UncraftedName:
Added Smiley's pattern to make Smiley smile. Interestingly, the pattern is the same for 5135 hl2 but the command doesn't do anything.

SmileyAG:
> 
> 
> Added Smiley's pattern to make Smiley smile. Interestingly, the pattern is the same for 5135 hl2 but the command doesn't do anything.

😃. I didn't found any function by using Search Memory with mine pattern in hl2 client.dll of 5135 build. 
изображение

YaLTeR:
Thanks

PR #179, opened 2021-09-25
Support of overlay functionality for SDK2013 mods

SmileyAG:
I don’t understand yet how to draw the main HUD in overlay mode, so draft PR, but I found right pattern for `VGui_Paint` and SPT HUDs now draws.

@lipsanen any thoughts?

изображение


lipsanen:
my thoughts: overlay stuff is hard

PR #186, opened 2021-12-21
Prevent double plugin_load

lipsanen:


spt/spt-serverplugin.cpp
```diff
@@ -262,6 +271,13 @@ bool CSourcePauseTool::Load(CreateInterfaceFn interfaceFactory, CreateInterfaceF
...
 void CSourcePauseTool::Unload(void)
 {
+	if (skipUnload)
+	{
+		// Preventing double load of plugin, do nothing on unload
+		skipUnload = false; // Enable unloading again
+		return;
+	}
+
 #if !defined(OE)
```
YaLTeR, line 281: Should this also set pluginLoaded = false?

lipsanen: No, when this occurs the plugin is still loaded and is running on the same instance of that DLL if that makes sense.

YaLTeR: I mean when you unload the last copy.

lipsanen: It still unloads and loads fine, in a real unload the library gets freed completely it seems.
image

YaLTeR: ah, cool

PR #187, opened 2021-12-26
Added ihud and tas_pause crash fix

evanlin96069:
New commands:
y_spt_ihud
y_spt_ihud_button_color
y_spt_ihud_font
y_spt_ihud_grid_padding
y_spt_ihud_grid_size
y_spt_ihud_shadow_color
y_spt_ihud_shadow_font_color
y_spt_ihud_x
y_spt_ihud_y

y_spt_ihud 1 for normal ihud
y_spt_ihud 2 for analog movement / view
ihud

Prevent load crash while tas_pause.
(Auto set tas_pause to 0 when loading)

### DIFF
--- a/spt/features/ihud.cpp
+++ b/spt/features/ihud.cpp
@@ -4,6 +4,8 @@
 
 #include "..\cvars.hpp"
 #include "playerio.hpp"
+#include "signals.hpp"
+#include "property_getter.hpp"
 #include "..\sptlib-wrapper.hpp"
 #include "Color.h"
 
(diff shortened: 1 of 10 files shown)


spt/features/ihud.cpp
```diff
@@ -0,0 +1,278 @@
...
+		{
+			ang /= len;
+			ang *= pow(len / 180.0f, 0.2);
+			ang.y *= -1;
+		}
+
+		// Movement
+		Vector movement = inputMovement;
+		int movementSpeed = movement.Length();
+		int maxSpeed = movementSpeed > 150 ? 450 : 150; // For Portal TAS
```
lipsanen, line 174: Include header ent_utils.hpp and use `utils::GetProperty<float>(0, "m_flMaxSpeed");` instead of hardcoding the maxspeed.

spt/features/playerio.cpp
```diff
@@ -412,6 +419,11 @@ void __fastcall PlayerIOFeature::HOOKED_CreateMove_Func(void* thisptr,
 
 	ORIG_CreateMove(thisptr, edx, sequence_number, input_sample_frametime, active);
 
+#if defined(SSDK2007)
+	auto cmd = reinterpret_cast<CUserCmd*>(pCmd);
+	spt_ihud.SetInputInfo(cmd->buttons, Vector(cmd->sidemove, cmd->forwardmove, cmd->upmove));
+#endif
+
```
lipsanen, line 426: This would be better if you added a new signal into spt/utils/signals.hpp/cpp, for example `Gallant::Signal1<CUserCmd*> CreateMove;` then you could connect to it inside ihud.

lipsanen (changes requested):

spt/features/playerio.cpp
lipsanen: For example see SV_ActivateServerSignal in generic.cpp

lipsanen:
looks very good, just one formatting thing to fix

lipsanen:
thanks!

PR #192, opened 2022-01-19
Improve y_spt_find_portals

UncraftedName:
In-game, `y_spt_find_portals` is used when there are multiple portals and the overlay vars cannot determine the right portal to use. The only info that `y_spt_find_portals` prints is the portal position, which is just balls. This is especially balls with evanlin's VAG trace commands, so I've added some extra info and changed it to use client-side ents (and print client-side ent indices as well).
unknown
unknown


lipsanen:
Thanks!

PR #193, opened 2022-01-20
Fixed overlay aspect ratio not adapting to the game screen

SirWillian:
The overlay renderer doesn't adjust itself properly to the game's aspect ratio (only noticeable once _y_spt_overlay_swap is used). This fixes that by using the engine's GetScreenAspect function

Example in 21:9 aspect ratio, toggling _y_spt_overlay_swap
Before:
before

After:
after

### DIFF
--- a/spt/features/overlay.cpp
+++ b/spt/features/overlay.cpp
@@ -59,6 +59,7 @@ void Overlay::InitHooks()
 {
 	HOOK_FUNCTION(client, CViewRender__Render);
 	HOOK_FUNCTION(client, CViewRender__RenderView);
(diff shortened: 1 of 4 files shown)


SirWillian:
Apparently the issue is that OverlayRenderer is guarded by `#ifndef OE`, and the Overlay feature uses `#if defined(SSDK2007)`. I suppose I can change one to match the other, I'm just not sure which case is better

lipsanen:
Use #IFDEF SSDK2007 imo, the overlay isn't supported on other engine versions.

lipsanen:
Thanks!

PR #206, opened 2022-04-17
Add y_spt_override_tpose

UncraftedName:
Inspired by https://github.com/MattMcNam/SetSequence (one of the plugins used in the making of _She Got Bored_). The SetSequence plugin doesn't appear to work - it references the virtual `C_BaseAnimating::SetSequence` function. But at least on 5135, that function is not virtual.


lipsanen:
tyvm

Because of various git shenanigans I had to cherry-pick this commit in order for the commit history not get completely mangled, hence why this is marked as closed and not merged

PR #215, opened 2022-06-22
Changed ihud x y offset from screen pixel to percentage.

evanlin96069:
Changed ihud x y offset from screen pixel to percentage.
Swapped ihud buttons x y (now x is horizontal, y is vertical).
Fixed some typos.


lipsanen:
thanks

PR #226, opened 2022-07-11
Physics shadow stuff and pattern formatting

UncraftedName:
I thought, since spinning is so fun, why not add a spin command? And then I did: `y_spt_set_shadow_spin <roll>`  (only implemented for 5135).
image
Currently it is only known how to make the shadow spin about the x axis (roll) which is why you don't specify three angles. You can also see the angles and position with `y_spt_hud_shadow_info`.
image
This gets into the evil realm of working with ivp units, so I added some very basic structures (that match the ones used by ivp) specifically for that purpose.

I also added a simple hack for making sure that you don't see Chell's brain every time you saveload in thirdperson; my brain has been splattered across the walls from working on this so I've seen enough. It looks like that doesn't happen in portal 3420, so I added the hack for 5135 + latest steampipe.

Finally, I moved every pattern to the feature file in which it's used, so now every feature looks something like this:
image
I think this is especially nice for patterns where the pattern index is used, as you can tell what the index is without having to look through a separate pattern file. This also had the benefit of decreasing the size of spt.dll by about 0.6MB and speeding up build times! (A full rebuild for me went from ~38s to ~24s). All unused patterns are in patterns_archive.cpp.

For the sole purpose of being able to have namespaces like `patterns::server`, I bumped the language version to C++17.


tmob03:
@UncraftedName why was this closed this feature would be very useful for bms

UncraftedName:
@tmob03 I reformatted it and added some other stuff here https://github.com/YaLTeR/SourcePauseTool/pull/229.

PR #232, opened 2022-07-21
Fix y_spt_additional_jumpboost and Add y_spt_aircontrol

evanlin96069:
- Fix y_spt_additional_jumpboost using 2838's implementation.
- Add FCVAR_DEMO flag to y_spt_autojump.
- Change the incorrect pattern label in FinishGravity.
- Add y_spt_aircontrol for Portal.

### DIFF
--- a/spt/cvars.hpp
+++ b/spt/cvars.hpp
@@ -12,7 +12,7 @@ void FormatConCmd(const char* fmt, ...);
 extern ConVar y_spt_pause;
 extern ConVar y_spt_motion_blur_fix;
 extern ConVar y_spt_autojump;
-extern ConVar y_spt_additional_jumpboost;
+extern ConVar y_spt_jumpboost;
 extern ConVar y_spt_aircontrol;
 extern ConVar y_spt_focus_nosleep;
 extern ConVar y_spt_stucksave;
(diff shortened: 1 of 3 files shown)


InputEvelution:
What exactly is "HL2DM movement mechanic" supposed to be about? I'm aware that autojump doesn't seem to work properly in that game, but I don't think the player ever gets a horizontal boost upon jumping.

evanlin96069:
> What exactly is "HL2DM movement mechanic" supposed to be about? I'm aware that autojump doesn't seem to work properly in that game, but I don't think the player ever gets a horizontal boost upon jumping.

tbh, I'm not sure if that's HL2DM movement mechanic because I didn't play HL2DM before... It's the movement you got if you set `maxplayers` > 1 in Portal or HL2. Basically, no jump boost, you can bhop.

lipsanen:
thanks

lipsanen:
> It may also be worth removing the additional_ from the command's name as well.
There was some discussion about the entire naming scheme in spt and I think most people agreed that it should only be spt_*, ie. drop the leading _y and y_. I probably would still want to mirror the old cvars so that they work by their old names, afaik that should be possible.

PR #233, opened 2022-07-24
Portal 3420 TAS support and more patterns

evanlin96069:
- 1910503 and 7197370 patterns for FinishGravity. (Now y_spt_jumpboost works on steampipe versions)
- ControllerMove patterns for 3420
- Use CheckJumpButton offsets in autojump for mv and player offsets in [strafestuff.cpp](https://github.com/YaLTeR/SourcePauseTool/compare/master...evanlin96069:3420?expand=1#diff-68301b67bcc384cdc01bcfdc24f1334446f221634143771bdc978fe0f85b9865) because different version can have different offsets
- Changed CPortalGameMovement__TracePlayerBBox patterns so it can also use in 3420
- Changed EndOfTeleportTouchingEntity patterns so it can also use in 3420


lipsanen:
thanks

PR #235, opened 2022-08-06
IVP RNG seeding

UncraftedName:
Despite previous attempts it looks like seeding IVP does have a very significant effect on physics RNG, but it may depend heavily on when the seeding happens. This change allows seeding during the load before any IVP simulation happens with `y_spt_set_ivp_seed_on_load`. Also, the seed is printed on every load which you can see with developer 1 - we can go back to old scripts, run them until they work, and set the seed for all future runs.

For afterframes scripts you would set the cvar before the load command.
For srctas scripts you can put the cvar on the settings line.

If there's some hook that spt already has that gets called during the load before any IVP simulation, (e.g. OnRestore happens way after) I can move the seeding there instead.

### DIFF
--- a/spt/features/rng.cpp
+++ b/spt/features/rng.cpp
@@ -9,16 +9,34 @@
 #include "usercmd.h"
 #endif
 
+ConVar y_spt_set_ivp_seed_on_load(
+    "y_spt_set_ivp_seed_on_load",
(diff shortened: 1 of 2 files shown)


lipsanen:
good stuff

PR #238, opened 2022-09-01
Fixed 3420 tas_pause crash and override fov.

evanlin96069:
- Fixed 3420 tas_pause crash by adding pHost_Frametime offset for 3420.
- `y_spt_cam_fov` now overrides all camera FOV, but won't record the FOV in demos. (For TASing/segmenting in 3420 with FOV 90 but still recording FOV 75 demos)


lipsanen:
IMO remove y_spt_cam_fov entirely and use _y_spt_force_fov here instead.

lipsanen:
thanks

PR #241, opened 2022-09-11
Fixed crash in SDK2013 modifications

SmileyAG:
Valve somehow fucked up a build system at SDK2013 Base, so if u launch some modification, it outputs `0` as build number.

At first I thought it was issue at my side, but it's not, I asked a few other runners and they have a same build output, so I have to do this hack to avoid crash:


lipsanen:
The build number for newer builds comes from the steam.inf file so if that's missing from the mod then yea it's probably 0

lipsanen:
tyvm

PR #246, opened 2022-09-28
Fixed autocomplete crash and refactor.

evanlin96069:
Fixed autocomplete crash when entering not existing directory.
Added double quote autocomplete.


lipsanen:
tyvm

PR #248, opened 2022-09-30
New rendering system!!!

UncraftedName:
Dear Jukspa & Yalter,
Please merge.
Thanks - UncraftedName

## Problem
The debugoverlay is extremely slow because it creates a single mesh for every little piece you wish to render. I initially came across this as a huge performance hit when rendering saveglitch collision visualization (even evanlin's VAG trace stuff is slow and it just draws a couple circles with 200 verts). I did implement a much faster system that hooked a function used by vcollide_wireframe (phys-vis). That system is *extremely* crude and limited however, and it doesn't allow rendering lines.

## Solution
Kidnap mlugg and create a rendering system designed from the ground up with performance and complex visualizations in mind. This will hopefully pave the way for more cool visualizations in the future and maybe possibly some UI stuff.

This system is meant to be a complete replacement for the debugoverlay (aside from text/hud stuff, which will hopefully be added in the future), it also allows us to add any drawing primitives that we desire. It is roughly broken up into two modules: mesh creation & rendering; the details of the interface/internals are still in the air and may get a rework in the future. I also went through and changed (almost) everything that uses the debugoverlay to use this new system.

## Changes
- Miscellaneous stuff:
  - some stuff in spt_entutils was not const that could have been
  - added a special ifdef* for tas_restart
  - any cvars that I used/changed are not included and/or removed from cvars.hpp**
  - added a signal to spt_overlay which I use for rendering
- Mesh creation/rendering:
  - y_spt_draw_overlay_examples: examples are cool
  - y_spt_draw_overlay_debug: debugging is cool
- Saveglitch collision visualization:
  - world geometry meshes are now cached which seems to give a somewhat significant performance bump
- Seamshot stuff:
  - this was spread out across approximately 5 morbillion files, it's all in one place now***
- VAG visualizations:
  - caches a bunch of maths instead of recalculating every frame
  - sped up & simplified some existing calculations
  - slowed down & over-complicated everything else by adding more visualizations
    - y_spt_draw_vag_lock_entry: you can now choose to keep the entry portal or exit portal fixed in place
    - y_spt_draw_vag_trace: this now shows you how the VAG destination will change as you move your free portal
  - changed all visualization commands to start with y_spt_draw

*spt's compile-time dependencies really really should be named if features depend on each other, I added names to ifdefs in tas_restart and the rendering system

**cvars.hpp is all but useless imo, it's outdated and almost no features use cvars from other features - you can just declare the cvar in whatever code you are using it in

***the seamshot stuff could be improved - it uses an offset instead of a pattern for a couple things and it probably shouldn't be using TraceFirePortal, I am considering making a better version of it

## Brief overview of the system
At the start of every frame, the renderer will call the overlay signal for users to "draw" their meshes. The user can create dynamic meshes in this signal or create static meshes whenever, then feed them to the renderer. The renderer queues all meshes until later during the frame - for every draw opaque/translucent call the game does, we must render our meshes. We can render static meshes directly, but dynamic meshes are recreated every time using the info the user gave us. We can also apply some primitive changes to meshes just before we render them (this is done by the user giving an appropriate callback). At the start of the next frame before we send the overlay signal again, the renderer and mesh builder will clear their internal buffers and the process begins again.
image
image
image
image


lipsanen:
cool stuff, thanks

mlugg:
> Kidnap mlugg and create a rendering system...

help i just want to see my family again

PR #249, opened 2022-10-04
Added autocomplete for existing commands.

evanlin96069:
Replace autocomplete functions for `playdemo`, `load`, and `exec` with better ones.
Changed tas_script_load help text.
Added autocomplete for tas_script_search.

### DIFF
--- a/spt.vcxproj
+++ b/spt.vcxproj
@@ -1131,6 +1131,7 @@
     <ClCompile Include="spt\feature.cpp" />
     <ClCompile Include="spt\features\afterframes.cpp" />
     <ClCompile Include="spt\features\aim.cpp" />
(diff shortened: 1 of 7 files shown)


lipsanen:
what's the situation with steampipe autocomplete, are these all supported on steam/does something still crash etc.

evanlin96069:
Steampipe:
- Replacing autocomplete functions for existing commands (like `playdemo`) won't crash.
- Autocomplete for the plugin's commands (like `tas_script_load`) will crash so it's currently disabled.

DMoMM:
- Autocomplete will crash so replacing autocomplete functions for existing commands is disabled.

HL2 OE (4044) works fine.
BMS build was not tested because I didn't own that game.

lipsanen:
gotcha, tyvm

PR #251, opened 2022-10-18
Add Entropy Zero 1 and 2 Autojump support

tmob03:


spt/features/autojump.cpp
```diff
@@ -239,6 +241,12 @@ void AutojumpFeature::LoadFeature()
 		case 18:
 			off_mv_ptr = 2;
 			break;
+		case 20:
+			off_mv_ptr = 2;
+			break;
+		case 21:
```
SmileyAG, line 247: `case 21:` is extra here, since u had 21 patterns in total (ptnNumber counting is starts from 0)

SmileyAG (changes requested):

spt/features/autojump.cpp
tmob03: oops yeah i had two sigs, one for ez1 and another for ez2 before i combined them

lipsanen:
tyvm

PR #253, opened 2022-10-22
Black Mesa 0.91 BBox Rotation Support

tmob03:


UncraftedName:
I'm pretty sure you want to have a wild card in there. Unless I'm mistaken on the disassembly the  `A1 D0 82 0D 10` should really be `A1 ?? ?? ?? ??` because of relocations.

lipsanen:
You wanna use the makesig.py script for generating the pattern in ghidra, I think the same thing exists in IDA if you are using that. It does all the byte masking correctly, so that your pattern doesnt randomly break between different launches of the game. Grab it from here https://github.com/YaLTeR/ghidra_scripts/blob/master/makesig.py, you need to put it in some ghidra script directory I forget.

lipsanen:
should probably add that in the contributing guide as well

tmob03:
@lipsanen Makesig.py makes this, isn't it too short?
`55 8B EC 81 EC E4 00 00 00 A1 ?? ?? ?? ?? 33 C5 89 45 ?? 8B 55 ??`

UncraftedName:
makesig creates the shortest unique pattern for that dll

lipsanen:
tyvm

PR #254, opened 2022-10-23
Fixed crash in Estranged: Act 1, made a few autojump patterns for matching in most of SDK2013 mods 

SmileyAG:
Before you might say that I have broken something, here is a list of steampipe games that autojump works with a new patterns:

- Half-Life 2 (Latest)
- Portal (Latest)
- Entropy Zero
- Entropy Zero 2
- BMS (Mod)
- BMS (0.9 Retail)
- Estranged Act 1
- Transmissions: Element 120
- ICE
- Map Labs: Half-Life Eternal


lipsanen:
tyvm

PR #256, opened 2022-11-01
Changed language standard to C++20

SirWillian:
Don't mind the original branch name. Meant to work on something else and ended up doing this
Not really a C++ pro, so maybe some of the changes may not be the most adequate, but a summary of the changes is:

- Changed platform toolset to v142 (VS2019) and language standard to stdcpp20
- Updated README installing instruction pointing to the new platform toolset
- Worked around toolset bug by adding a new header (typeinfo.h)
  - Wasn't sure if it was better to make a new folder for it or just use one of the existing include paths
- Added compiler option to prevent PCH compilation from running out of memory
- Updated SPTlib submodule (perhaps wasn't 100% needed, but figured it was a decent idea) and SDK/episode1 submodule (required)
- Pointed to a different libMinHook vcxproj (so the previously used project could be kept on v141_xp)
- Made many variables const when possible
- Included some headers here and there as part of troubleshooting
  - Seems like some files in the project aren't self sufficient when it comes to which headers they include, but it all happens to work because the required headers are imported externally
- Moved some functions around in string_utils because it seemed like the compiler was choosing the wrong overloads and throwing errors
  - Maybe it would also magically fix itself once other errors in the project were solved but doesn't hurt to keep it as it is I guess
- Other misc/small changes required because of the language standard change

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,10 @@ A plugin for all your pausing needs.
     * *Loading SourcePauseTool more than once will crash the game!*
 
 ### Building
-0. You will need [Visual Studio 2019](https://visualstudio.microsoft.com/vs/older-downloads/) and [git](https://git-scm.com).
+0. You will need [Visual Studio 2019](https://visualstudio.microsoft.com/free-developer-offers/) (or later) and [git](https://git-scm.com).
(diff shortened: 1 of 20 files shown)


lipsanen:
tyvm

PR #261, opened 2022-11-15
HUD/Overlay/mesh rendering steampipe support

UncraftedName:
Changes:
- the HUD, overlay, & mesh rendering now work on steampipe
- HUD:
  - added a dedicated include guard `(SPT_HUD_ENABLED)`
- overlay:
  - added a dedicated include guard `(SPT_OVERLAY_ENABLED)`
  - removed overlay folder, simplified/merged stuff into the feature but kept portal_utils.hpp/cpp in the utils folder
  - added a dedicated include guard to portal_utils.hpp `(SPT_PORTAL_UTILS)`
- Mesh rendering:
  - added patterns for hl2
  - added AddCross and AddCPolyhedron
  - refactored everything too many times and added a diagram that shows the structure of the system (mesh_defs_private.hpp)

The way that the overlay used to work was by calling `CViewRender::Render()` twice which would in turn call `CViewRender::RenderView()` twice. Steampipe added a lot of VR related code into both of those functions, so for whatever reason calling `Render()` twice is not the way to go. It turns out there's a dedicated function for an overlay though - `QueueOverlayRenderView()`. With a bit of adjusting that seems to mostly works as-is, although it seems like the HUD has to be rendered for both the main/overlay view at the same time.

Not everything works on steampipe e.g. draw seams, vel hud, and oob hud don't load, and hops hud doesn't display correctly after running tas restart.

Speaking of tas restart, I have a dedicated system for handling that for static meshes - which is to keep a linked list through all of them and destroy them on feature unload. That could probably be abstracted away and done if needed for other structures.

Also bill and I were thinking that it'd be a good idea to remove signals.hpp altogether and simply have each signal in its respective feature. For example - I set this up with mesh rendering. It works similar to the hud callback system, but instead of calling `spt_hud.AddCallback(...)` I have `spt_meshRenderer.signal.Connect(...)`.

dodecahedron with AddCPolyhedron: 
image


spt/features/restart.cpp
```diff
@@ -17,22 +16,28 @@ CON_COMMAND(tas_restart_game, "Restarts the game")
...
+class RestartFeature : public FeatureWrapper<RestartFeature>
 {
-#ifdef SPT_TAS_RESTART_ENABLED
-	return true;
-#else
-	return false;
+protected:
+	virtual bool ShouldLoadFeature()
+	{
+#if defined(OE) || defined(SSDK2013)
```
lipsanen, line 25: What's the reasoning behind changing this check from ifdef SPT_TAS_RESTART_ENABLED ?

lipsanen:
For tas restart in general, the idea was that you would put everything that needs to be reset as a member to a feature, and only globals get preserved between restarts. Therefore if you wanna do something like a linked list, you can use std::list as a member to a feature which will clean itself up automatically.

spt/features/restart.cpp
lipsanen: idk how this is related to the header. Also idk if you are aware but include guard is a specific term used for definitions that prevent including the same header once, ie.
```
#ifndef XD
#define XD


// some header stuff

#endif```

**Comment by UncraftedName** [comment]
I wanted to stick with the static variables pattern because I think it makes the code a bit more readable. So I have a lot of code that looks like:
```c
BEGIN_TEST_CASE("Timmy", VEC_WRAP(0, -300, 0))
{
	static StaticMesh timmyMesh;
	if (!timmyMesh.Valid())
	{
		// make timmy if he needs to be recreated
		timmyMesh = MB_STATIC(mb.AddBox(vec3_origin, mins, -mins, a, c););
	}
	...
END_TEST_CASE()
```
After too much discussion with bill & mlugg, having a feature store the linked list instead is equivalent to what I have now (so long as each static mesh has an iterator into that list so it remove itself). But since it's not actually any different I don't think it's worth to change.

spt/features/restart.cpp
UncraftedName: yeah I realized that include guard isn't really the right word for it. I removed restart.hpp because that's where I previously defined `SPT_TAS_RESTART_ENABLED`. But I only used that in the mesh renderer and I don't need to anymore, so `SPT_TAS_RESTART_ENABLED` isn't needed for the rest of spt.

lipsanen: gotcha

lipsanen:
tyvm

PR #263, opened 2022-11-18
Steampipe tas_restart fixes and other things

UncraftedName:
Fixes:
- evanlin: fix `y_spt_cam_control` being borked with thirdperson
- remove possible crash in sg-collide-vis
- enable some hud vars for steampipe (that I forgot to enable previously)
- fix spt_playerio & property_getter breaking after using tas_restart
- fix `DrawFlagsHud()` behavior for mutually exclusive flags, this fixes `y_spt_hud_moveflags` and `y_spt_hud_movecollideflags` not displaying on 5135 & steampipe

Changes:
- remove `y_spt_calc_relative_position`
- move property getter to a feature so that it can unload
- `y_spt_find_portals` now tells you if there are no portals instead of doing nothing
- rename `spt_entutils` to `spt_entprops` - it lives in a file called ent_props and we already have utils/ent_utils, kind of an L on that naming scheme
- added patterns for `y_spt_hud_oob`

I *think* all the HUD vars are least available (and don't crash) on steampipe. Methinks we should use the SDK for tracing, but we may have to include `common\shared\headers\util_shared.h` and possibly parts of the corresponding cpp file for full functionality. I may look into this further if/when I work on draw_seams v2 again.


lipsanen:
tyvm

PR #264, opened 2022-11-19
Less warnings

UncraftedName:
Rebuilding on Release gave 143 warnings, now it gives 1 (from SPTLib).
- renamed some params/local vars to remove "hides class member" warning
- removed a couple signed/unsigned mismatches
- suppressed SDK imesh.h warning (operator+ deprecated between enumerations of different types)
- suppressed inline asm warning
- re-added the link against vstdlib.lib so I don't have to manually create KeyValuesSystem()
- globally suppressed enum forward declaration type warning, the SDK forward declared enums a lot so I don't think a full fix is feasible to PR into the SDK

Building for OE still gives some conversion warnings from ssemath.h and some uninitialized warning from vector.h. Not sure what the best fix is since those files are included everywhere and I think the latter could actually be useful.

### DIFF
--- a/spt/features/visualizations/renderer/source files/materials_manager.cpp
+++ b/spt/features/visualizations/renderer/source files/materials_manager.cpp
@@ -4,29 +4,7 @@
 
 #ifdef SPT_MESH_RENDERING_ENABLED
 
-#include "vstdlib\IKeyValuesSystem.h"
 #include "interfaces.hpp"
-
-// We need KeyValues to make materials; they in turn need KeyValuesSystem(), but we don't link against vstdlib.
-// Here's a little hack to make that work - find the KeyValuesSystem function by searching for its symbol.
-#pragma warning(push)
-#pragma warning(disable : 4273)
-IKeyValuesSystem* KeyValuesSystem()
-{
-	static IKeyValuesSystem* ptr = nullptr;
-
-	if (ptr != nullptr)
-		return ptr;
-
-	void* moduleHandle;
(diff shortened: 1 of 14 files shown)


lipsanen:
everything else looks fine apart from the fact that vstdlib.lib randomly got added as a dependency

spt/features/visualizations/renderer/source files/materials_manager.cpp
```diff
@@ -4,30 +4,8 @@
...
 #ifdef SPT_MESH_RENDERING_ENABLED
 
-#include "vstdlib\IKeyValuesSystem.h"
 #include "interfaces.hpp"
 
-// We need KeyValues to make materials; they in turn need KeyValuesSystem(), but we don't link against vstdlib.
-// Here's a little hack to make that work - find the KeyValuesSystem function by searching for its symbol.
-#pragma warning(push)
-#pragma warning(disable : 4273)
-IKeyValuesSystem* KeyValuesSystem()
```
UncraftedName, line 14: I did mention that as part of the changes above - I need to use the KeyValues system to make materials (from the SDK), but internally that calls `KeyValuesSystem()` from within the SDK. That function only exists in vstdlib.lib, otherwise I had to patch it back in by searching for its symbol in the dll. The problem with that approach is that you get a linker warning about an externally defined symbol which I could not figure out how to suppress with pragmas.

lipsanen: gotcha

PR #266, opened 2022-11-23
Add OE support or HUD.

evanlin96069:
HUD now works in OE (except DMoMM).
Remove some OE ifdefs in ent_props/ent_utils so I can property_getter for ihud.
Fix incorrect warning in y_spt_pause.


lipsanen:
tyvm

PR #269, opened 2022-11-27
Port of waezone's Microsegmenting

tmob03:
Uses only spt_ prefix, let me know if this change is still far out and i should make it y_.

Waezone's [microsegmenting](https://github.com/waezone/Random-Assortment-Of-Scripts/tree/master/Microsegmenting) is an ahk file designed to speed up segmenting, and allow for shorter segments overall as you don't need to manually edit a cfg file everytime. I have to the best of my ability ported this over. If someone knows of a way to make this less complicated for the end user let me know. Current a segment cfg would like this with it (things in [] are optional):

```
y_spt_autojump 1
y_spt_pause 1

bind "u" "spt_microsegmenting_record_segment [alias]"
bind f1 "spt_microsegmenting_load_segment [alias] [person]"
bind z "unpause; sensitivity 2.9"
bind mouse4 "spt_microsegmenting_stop_segment [alias]"
bind - "spt_microsegmenting_previous_segment"
bind = "spt_microsegmenting_next_segment"
[spt_segmenting_folder bmdq]
```

### DIFF
--- a/microsegmenting.cpp
+++ b/microsegmenting.cpp
@@ -1,13 +1,13 @@
 #include "stdafx.h"
 #include "..\feature.hpp"
 #include "..\sptlib-wrapper.hpp"
+#include <sstream>
 
 // Port of waezone's microsegmenting AHK file
 class MicroSegmenting : public FeatureWrapper<MicroSegmenting>
 {
 public:
 	int currentSegment = 1;
-	int previousSegment = 0;
 	std::string runFolder = "";
 	ConVar* playerName = nullptr;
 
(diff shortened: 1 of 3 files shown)


lipsanen:
Dont think the december release is gonna have the y_ removal yet, have still some work to do on that.

lipsanen:
you could change everything to y_ for the time being

lipsanen:
As for making it less complicated, I dont see a way to automatically start working on the current segment, ie. do something like `y_spt_microsegmenting_resume `, I don't really see how you can resume work after restarting the game, but I have never used this system so maybe it's not a relevant issue. The _segment postfix at the end of the commands also seems redundant.

evanlin96069:
Should probably use `cl_mouseenable 0` instead of `sensitivity 0` since sensitivity has the minimum value of 0.0001f (and show the default command in help string)
In my opinion, runFolder and currentSegment can just be a cvar and there’s no need for previousSegment since it seems always be currentSegment - 1

tmob03:
> Should probably use `cl_mouseenable 0` instead of `sensitivity 0` since sensitivity has the minimum value of 0.0001f (and show the default command in help string) In my opinion, runFolder and currentSegment can just be a cvar and there’s no need for previousSegment since it seems always be currentSegment - 1

cl_mouseenable does not work in all games, for example, hls and hdtf.

And i'll change those into a cvar

tmob03:
Actually change of plans idk how to do that 😐

PR #271, opened 2022-11-29
Remove TraceFirePortal using non-test shot

evanlin96069:
If are you near the portal, TracePortalFire will return 0.0265f (Near existing portal) only if the shot is not test shot,
otherwise, it will return 0.027f (Overlaps existing portal).
However, it will cancel the traveling portal shot if you look at a portal near the player before it lands.

In short, it might cause more unexpected glitches and the benefits of using it aren't much so I decide to remove it.


lipsanen:
tyvm

PR #274, opened 2022-12-06
Add various TAS tool features

evanlin96069:
- `tas_aim_pos <x> <y> <z>` tas_aim at a position
- `tas_aim_ent <id>` tas_aim at an entity (using entity index)
- `+y_spt_spam <key>` spam a key (attack, jump, duck, forward, back, use, left, right, moveleft, moveright, attack2, reload, speed, walk, zoom)
- Remove `+y_spt_duckspam`
- Use spam flag in framebulk field1 is now working
- DMoMM will now use a different framebulk action field (`jdu12wsklrx`). This change only affects DMoMM.
  + `jdu12` are the still the same
  + DMoMM don't use `+relaod`
  + `w`: `+speed` Walk in DMoMM, they don't use `+walk`
  + `s`: `+sprint` DMoMM sprint
  + `k`: `+kick`
  + `l`: `+leanleft`
  + `r`: `+leanright`
  + `x`: `+xana` special ability

- Removed unnecessary header files


spt/features/aim.cpp
```diff
@@ -177,6 +205,58 @@ CON_COMMAND(tas_aim, "Aims at a point.")
...
+{
+	if (args.ArgC() < 4)
+	{
+		Msg("Usage: tas_aim_pos <x> <y> <z>\n");
+		return;
+	}
+
+	int frames = -1;
+	if (args.ArgC() == 5)
+		frames = std::atoi(args.Arg(4));
```
lipsanen, line 218: frames argument here isn't included in the help text

spt/features/aim.cpp
```diff
@@ -177,6 +205,58 @@ CON_COMMAND(tas_aim, "Aims at a point.")
...
+		spt_aim.viewState.timedChange = false;
+	}
+	else
+	{
+		spt_aim.viewState.timedChange = true;
+		spt_aim.viewState.ticksLeft = std::max(1, frames);
+	}
+}
+
+CON_COMMAND(tas_aim_ent, "Aims at a entity. Usage: tas_aim_ent <id> [ticks]")
```
lipsanen, line 234: this might be more useful in hl2 if you could aim at entity + offset, just aiming at abs origin would aim at an NPC's feet for example

spt/features/playerio.cpp
```diff
@@ -512,25 +501,84 @@ void PlayerIOFeature::OnTick()
...
+{
+	if (!interfaces::engine)
+		return;
+	ArgsWrapper args;
+	SpamUp_wrapper(args);
+}
+static void SpamUp_wrapper(ArgsWrapper args)
 #else
-static void DuckspamUp(const CCommand& args)
+static void SpamUp(const CCommand& args)
```
lipsanen, line 567: the ifdef spam here looks a bit dirty, feel like this could be replaced with a macro

spt/features/aim.cpp
evanlin96069: frames argument means the ticks.

tmob03:
Can you keep duckspam for backwards compat

evanlin96069:
> Can you keep duckspam for backwards compat

Duckspam is back now. Good that I have created macros for declaring +/- action commands:)

spt/features/playerio.cpp
```diff
@@ -283,25 +284,14 @@ int __fastcall PlayerIOFeature::HOOKED_GetButtonBits_Func(void* thisptr, int edx
...
-			{
-				duckPressed = true;
-				rv |= (1 << 2); // IN_DUCK
-			}
-		}
-		else
-			duckPressed = false;
+		static int keyPressed = 0;
+		keyPressed ^= spamButtons;
+		rv |= keyPressed & spamButtons;
```
lipsanen, line 289: If I read this correctly this has kinda weird behavior, when you unpress spam for a certain key you have one of two cases: either the flag is set in keyPressed or it's not. This state is maintained until the spam is activated again. If the flag is set, then when the spam is next activated the key will be pressed on the second tick. If the flag is not set, then the key will be pressed on the first tick.

I think this would work the same every time:
```
static int keyPressed = 0;  
keyPressed = (keyPressed ^ spamButtons) & spamButtons;  
rv |= keyPressed;  
```

lipsanen:
tyvm

PR #275, opened 2022-12-14
More entity info for sg-collide-vis

UncraftedName:
I went through and cleaned up a little bit of some mesh rendering stuff (again), and restructured sg-collide-vis so that it's not just one huge function. I also added some functionality to `y_spt_draw_portal_env_ents` to draw entities far away from the portal.
image


lipsanen:
tyvm

PR #277, opened 2022-12-27
Disable Vcpkg

thisis2838:
Some users (like me) has an option to enable Vcpkg, which if enabled, will cause all SPT builds to have an external absolute path reference to a minHook .dll. This causes the plugin to only be loadable if such the minHook dll exists in that specific path.


lipsanen:
tyvm

PR #278, opened 2023-01-02
Change precompiled headers setup to fix IntelliSense error

thekovic:
I got the precompiled headers working with no IntelliSense error while having the file be named stdafx.hpp. Here's what I did:
In SPT Project Properties->Configuration Properties->C/C++->Precompiled Headers, I changed the name of the precompiled header file to stdafx.hpp.
Then, I moved stdafx.hpp and stdafx.cpp to somewhere so that all source files in all configurations can include from it using the same statement. That folder turned out to be /spt/utils (could be anything else if you bother changing the include setup).
Finally, just as a small technical note, I had to make sure that stdafx.cpp in its new location is being used to create the .pch file.

Now, all files can use `#include "stdafx.hpp" to use precompiled headers while not generating an error in Visual Studio's IntelliSense.

Because this pull request includes major changes to the spt.vcxproj file, it also happens to include my change of retargetting the solution to MSVC v143 because I needed to do that on my PC in order to compile SPT. If this is problematic, I can force push a change to revert this retargetting.


lipsanen:
The .github/workflows/CI.yml file probably needs some work

thekovic:
> The .github/workflows/CI.yml file probably needs some work, looks like it's missing the v143 build tools

I tried changing the CI to `runs-on: windows-2022` (https://github.com/thekovic/SourcePauseTool/commit/111e8847b5b93dd6c3b8087eb693dd3ecaa3b89c) and that got Debug builds working in CI but the Release builds throw an error `LINK : fatal error C1047: The object or library file 'D:\a\SourcePauseTool\SourcePauseTool\lib\Release\libMinHook.x86.lib' was created by a different version of the compiler than other objects like '.\Release BMS\Hooks.obj'; rebuild all objects and libraries with the same compiler [D:\a\SourcePauseTool\SourcePauseTool\spt.vcxproj]`

Do you think it's better to solve this by fixing this error (which I assume is on the SPTLib side) or should I revert spt.vcxproj to target MSVC v142 again?

lipsanen:
looks about the same, there's one 3 minute build but probably just random

lipsanen:
tyvm

PR #279, opened 2023-01-02
New features & Black Mesa hud text support

thisis2838:
New features:
- Added `_y_spt_afterticks` which is a `host_tickcount`-based `_y_spt_afterframes` alternative
- Added `y_spt_saveloads` which does automated save/loading
- Added `y_spt_noclip_nofix` which disables position fixing when disabling noclip.
- Added `y_spt_portal_no_ground_snap` which disables an extra check in CategorizePosition that causes sticky noclip
- Added `y_spt_free_oob_movement ` which adds free OOB movement

Code changes:
- Added macros for getting modules, finding multi-match patterns, and working with direct byte replacements.

Black Mesa:
- Added hud text support
- Added `y_spt_bms_flashlight_fix` which removes extra movement from the flashlight.


spt/features/bms_flashlight_fix.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#ifdef BMS
+
+static void BMSFlashLightFixCVarCallback(IConVar* pConVar, const char* pOldValue, float flOldValue);
+
+ConVar y_spt_bms_flashlight_fix("y_spt_bms_0point9_flashlight_fix",
+                                "0",
+                                0,
+                                "Disables all extra movement on flashlight. (0.9 ONLY, FOR NOW!!!)",
+                                BMSFlashLightFixCVarCallback);
```
lipsanen: Warnings about this command not working on other versions are redundant as versions where it doesnt work the command should get deleted.

spt.vcxproj
```diff
@@ -279,7 +282,9 @@
     </Bscmake>
     <PostBuildEvent>
       <Message />
-      <Command>copy "$(TargetPath)" "D:\Games\Source Unpack\portal\"</Command>
+      <Command>md "$(IntDir)..\..\builds"
+del "$(IntDir)..\..\builds\$(TargetFileName)"
+copy "$(TargetPath)" "$(IntDir)..\..\builds\$(TargetFileName)"</Command>
```
lipsanen: where does this put the dll?

spt/features/autojump.hpp
```diff
@@ -9,6 +9,40 @@
...
+	    "missinginfo1_4_7",
+	    "55 8B EC 83 EC 44 56 89 4D D0 8B 45 D0 8B 48 08 81 C1 ?? ?? ?? ?? E8 ?? ?? ?? ?? 0F B6 C8 85 C9",
+	    "te120",
+	    "55 8B EC 83 EC 20 56 8B F1 8B ?? 04 80 ?? 48 ?? 00 00 00 74 0E 8B ?? 08 83 ?? 28 02 32 C0 5E 8B E5 5D C3",
+	    "BMS-Retail-3",
+	    "55 8B EC 83 EC 0C 56 8B F1 8B ?? 04 80 ?? ?? ?? 00 00 00 74 0E 8B ?? 08 83 ?? 28 02 32 C0 5E 8B E5 5D C3",
+	    "missinginfo1_6",
+	    "55 8B EC 83 EC 1C 56 8B F1 8B 4E 04 80 B9 ?? ?? ?? ?? ?? 74 0E 8B 76 08 83 4E 28 02 32 C0 5E 8B E5");
+}
+
```
lipsanen, line 45: Don't move the patterns in to the header and don't search for the pattern in two separate places. Put the headers in one cpp file and put the result in a public member if you need to use it somewhere else.
This is because:

1. MSVC sucks and can't dead code eliminate the patterns away for any cpp file that includes this header. This inflates the binary size of the dll.
2. Searching for the pattern twice is slower than searching it once.

spt/features/noclip_nofix.cpp
```diff
@@ -92,39 +92,42 @@ void NoclipNofixFeature::LoadFeature()
...
 			jmpDistance = jmpOff + 1;
+			goto success;
 		}
-	}
 
+	}
+	return;
+	
+	success:;
 	InitConcommandBase(y_spt_noclip_nofix);
```
lipsanen, line 109: you could just put the InitConcommandBase in the inner loop here and get rid of the goto

spt/features/portal_nogroundsnap.cpp
```diff
@@ -0,0 +1,112 @@
...
+			continue;
+
+		INIT_BYTE_REPLACE(Jump, start - i + 2);
+		goto success;
+	}
+
+	return;
+
+	success:
+	InitConcommandBase(y_spt_portal_no_ground_snap);
```
lipsanen, line 94: Could put the InitConcommandBase inside the for loop here as well

spt/features/afterticks.cpp
```diff
@@ -13,18 +13,18 @@ static bool afterticksPaused = false;
 static int afterticksDelay = 0;
 
 ConVar _y_spt_afterticks_await_legacy("_y_spt_afterticks_await_legacy",
-                                       "0",
-                                       FCVAR_TAS_RESET,
-                                       "Set to 1 for backwards compatibility with old scripts.");
+									   "0",
+									   FCVAR_TAS_RESET,
+									   "Set to 1 for backwards compatibility with old scripts.");
```
lipsanen, line 18: formatting

spt/features/saveloads.cpp
```diff
@@ -115,10 +115,10 @@ void SaveloadsFeature::LoadFeature()
...
 void SaveloadsFeature::Begin(int type_,
 							 const char* segName_,
-                             int startIndex_,
-                             int endIndex_,
-                             int ticksToWait_,
-                             const char* extra)
+							 int startIndex_,
+							 int endIndex_,
+							 int ticksToWait_,
+							 const char* extra)
```
lipsanen, line 121: formatting

lipsanen:
there appears to be some conflicts with the master branch

spt.vcxproj
lipsanen: ah in the builds/ folder

spt/features/free_oob.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "convar.hpp"
+#include "autojump.hpp"
+#include "interfaces.hpp"
+
+static void FreeOOBCVarCallback(IConVar* pConVar, const char* pOldValue, float flOldValue);
+
+ConVar y_spt_free_oob_movement("y_spt_free_oob_movement",
+						      "0",
+						       0,
+                              "Disables noclip's position fixing.",
```
evanlin96069, line 12: Wrong help string? (It’s the same as `y_spt_noclip_nofix`)

spt/features/autojump.hpp
thisis2838: ORIG_CheckJumpButton points to the relocated original bytes by the time any other feature has access to it, if using HOOK defines.

spt.vcxproj
thisis2838: Figured it was cleaner that way since you can use symlinks to get it anywhere else.

thekovic:
I'd personally prefer calling it `y_spt_free_void_movement` because calling it Out of Bounds can be misinterpreted.

spt/features/saveloads.cpp
```diff
@@ -0,0 +1,266 @@
...
+Please load the save from which save/loading should begin.\n\n------\n",
+	    _saveloadTypes[type_],
+		segName_,
+		startIndex_,
+		endIndex_,
+		endIndex_ - startIndex_ + 1,
+		ticksToWait_,
+		this->extraCommands.c_str());
+
+	EngineConCmd("bind L y_spt_saveloads_stop");
```
lipsanen, line 180: I thought I commented on this already but this hardcoded bind should probably be deleted

thisis2838: You haven't but what should I do instead to communicate that the process can be stopped midway?

lipsanen: another option would be to make it a toggle, like +y_spt_saveloads and when you stop holding it you stop saveloading

thisis2838: I don't think that's really optimal because many save/loads are long and people leave them running in the background while they alt tab to something else (with y_spt_focus_nosleep)

lipsanen: That could be solved via aliases but then again you could also create a +/- alias for this. Either way is fine but the hardcoded bind has gotta go

spt/features/autojump.hpp
lipsanen: you can check the pointer during PreHook, it happens between finding patterns and hooking as described in the contributing guide

spt/features/saveloads.cpp
thisis2838: I can't create a +/- for this since it begins once you load a save

thisis2838: Fuck it I'll just put up a warning I'm tired of waiting for a reply

lipsanen:
tyvm

PR #280, opened 2023-01-13
""""""Fix"""""" Black Mesa hardlocking

tmob03:
As far as I can tell this [function ](https://github.com/VSES/SourceEngine2007/blob/43a5c90a5ada1e69ca044595383be67f40b33c61/src_main/engine/baseserver.cpp#L2138)does nothing important for singleplayer, so its fine to just never run, but I also suck at reading code sooooo.

### DIFF
--- a/hardlock.cpp
+++ b/hardlock.cpp
@@ -1,0 +1,41 @@
+#include "stdafx.hpp"
+
+#ifdef BMS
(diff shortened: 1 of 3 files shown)


lipsanen:
tyvm, if it breaks something I'll trust you'll fix it later

PR #281, opened 2023-01-16
improve bms version detecting

tmob03:


tmob03:
what why is oe failing the error is in not in a compiled in part

tmob03:
whatever

lipsanen:
is there a problem with version detection? In general I think spt has too many ifdefs already

tmob03:
Smaller DLL size is what I was going for. As for the BMS changes, I changed those because Mod and current release were calling the function for 0.9 detection

lipsanen:
I would prefer no added ifdefs. The BMS functions are kinda weirdly named, you would think DoesGameLookLikeBMS would match all versions but it doesnt

lipsanen:
tyvm

PR #283, opened 2023-01-16
y_spt_mousemove & Correct BMS Retail VEngineClient

thisis2838:
- Black Mesa Retail broke cl_mousemove and cl_mousenable in terms of using them to toggle viewangle changing.
- The VEngineClient version seen in the Black Mesa SDK is older than the current Retail versions used by runners.


lipsanen:
tyvm

PR #284, opened 2023-01-17
Fastloading and plugin callback signals

tmob03:
As draft because there is an issue about needing to hook disconnect, or at least find a way to know when player is going to the menu, because if you don't it will set mat_norendering to 1 when on the menu. As of now I'm completely clueless as to how to do that.


lipsanen:
tyvm

PR #287, opened 2023-01-25
Portal placement grid

UncraftedName:
Added the following commands/cvars:
- `y_spt_draw_pp_grid`
- `y_spt_draw_pp_grid_fov`
- `y_spt_draw_pp_grid_ms_per_frame`
- `y_spt_draw_pp_grid_type`
- `y_spt_draw_pp_grid_width`

This uses TraceFirePortal to trace thousands of rays from the player's eyes and check where a portal is possible to place. It can be used to find fizzle spots or tricky to place portals.
2023-01-25_11-08-21_1280 gif

I improved evanlin's `y_spt_draw_pp` stuff to also show fizzle spots.

To make some of this feature work, I implemented a (server-side) trace filter that should function identically to what the portal gun uses, so unless we actually care about portal placement we can simply call `engineTraceServer->TraceRay()`. This should be faster and doesn't rely on the portal gun existing. I updated draw_seams to use this instead of `TraceFirePortal()`.


lipsanen:
tyvm

PR #291, opened 2023-02-16
Fix and enhance camera control

evanlin96069:
Screenshot 2023-02-15 232658

- Add `y_spt_cam_drive_speed`: Speed for moving in camera drive mode
- Add `y_spt_cam_path_interp`: More interpolation type

- Fix incorrect time offset when playing cinematic mode
  - Fix calculating the time offset at incorrect place
  - Fix hooking incorrect StartPlayBack in 5135
- Better looking interpolation path render
- Use MeshBuilder to render interpolation path
- Use PostButtonPressed/ReleasedEvent to block inputs
  - The old one uses CreateMove which will block all inputs including TAS inputs
  - The new one only blocks keyboard inputs, so now it's possible to use drive mode while playing the TAS script
- Rename some variables to camelCase
- `y_spt_cam_path_setkf`, `y_spt_cam_path_rmkf`, and `y_spt_cam_path_clear` can now be use when not playing demo.

Known problems:
- PostButtonPressed/ReleasedEvent cannot block mouse button inputs (mouse1~5, mwheelup/down)
- Ticks for interpolation path are still using debugOverlay. Those texts can sometimes stuck on the screen when `y_spt_cam_path_draw 0`.

### DIFF
--- a/spt/features/camera.cpp
+++ b/spt/features/camera.cpp
@@ -10,6 +10,7 @@
 #include "spt\utils\interfaces.hpp"
 #include "spt\utils\signals.hpp"
 #include "spt\cvars.hpp"
+#include "visualizations/renderer/mesh_renderer.hpp"
 
 #include "usercmd.h"
 #include "view_shared.h"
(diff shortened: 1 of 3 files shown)


lipsanen:
tyvm

PR #294, opened 2023-02-24
Fix crash for 7122284 and assert for mesh sorting

UncraftedName:
xeonic wanted spt for an ancient portal 1 version lol


lipsanen:
tyvm

PR #295, opened 2023-02-26
Mesh renderering refactor (again!)

UncraftedName:
This is part 1 of the changes necessary to get text working, note that I **haven't** actually added text rendering yet. The reason that it's just part 1 is mostly because it has some important changes to the user interface of the mesh system and there have been several additional visualizations since I started working on this (2 months ago), so I mostly want to avoid merge conflicts.

Most of the important changes are summarized in the commit messages but I'll describe a couple things here.

I added caching to the `DoesGameLookLike` functions because I think I will end up using those thousands of times a tick, and as far as I know the `g_pCVar->Find` functions look through the entire linked list of cvars/commands.

The main change to the mesh system is to allow it to work with arbitrary materials (which is something I'll need for glyph atlases). This feature required major changes so I figured I would try to organize the whole system a tiny bit better while I was at it. Notably, I changed the user interface to allow the z-testing flag to be passed to every mesh component (before it was per-mesh). I also renamed everything and split it all up into a few more files.

Due to the internal structure now the system appears to be significantly slower, but most of that is cancelled out with the mesh fusing optimization that I added which can be seen with `y_spt_draw_mesh_debug 2`.
image

Due to the caching system I used with `y_spt_draw_portal_env_ents`, I had to make some big changes to that feature as well. I wanted to add more colors to this feature anyways, so I modified that feature as part of the same commit. These additional colors describe some of the different collisions you can have with entities when you have saveglitch.
image


lipsanen:
tyvm

PR #297, opened 2023-03-10
SPT Revolution

SirWillian:
Finally got this mildly done. Still have to adjust the help texts of commands but I'll do that last to make looking through this a bit easier. Main changes to look at:
- Commands now always start with "spt_" in-game (big overhaul to **_spt/cvars.cpp_**)
  - Before registering, the list of unregistered commands is checked to add new commands with new names linked to the old commands (which means that we don't have 90 altered files to get back-compat done). The old commands are then marked with FCVAR_HIDDEN so they don't show up in autocomplete.
    - ConVars are linked to the original to the originals through the ConVarProxy class, which changes just the bare minimum to get a new ConVar registered that actually links to another ConVar
    - ConCommands just reuse the command callback from the original command
  - Since OE doesn't have FCVAR_HIDDEN natively, that was also reimplemented for autocomplete purposes in the _**CvarStuff feature**_
    - Technically that flag also affects the find command but I didn't bother with that
    - Maybe this should be in the Autocomplete feature but that one is meant for autocompleting existing commands. Could repurpose it instead potentially
  - OE doesn't have a different cvar registering process now. For that, some SDK differences were reimplemented whenever needed
- Fused **_spt/utils/command.hpp_** with **_**spt/utils/convar.hpp**_**
  - There was already some weird cross-dependency between these headers anyway and command.hpp just added some autocomplete stuff, which made sense to be merged with convar.hpp
- Cleaned _**_spt/cvars.hpp_**_ to keep only ConVars defined by the game
  - No one knew what to do with the 100 lines of externs declared there, so I just got rid of them and moved the extern declarations to the files that needed them
  - There are some uses of game-defined ConVars in some places that should be moved here but that can be done later
- Small changes to the AutoCompleteList classes so autocomplete objects are compatible with multiple commands at once and to simplify the logic a little bit


SirWillian:
Changes made beside the string changes: [diff](https://github.com/SirWillian/SourcePauseTool/compare/568cae8888e99605c9433994c0cb850173f88806..9d77c86126ec0dcd4be1352041ed0b447f9ce13c)

lipsanen:
tyvm

PR #298, opened 2023-03-19
Disable autocomplete feature in Black Mesa

thekovic:
Autocomplete feature is broken in Black Mesa (0.9 and Definitive Edition) so I decided to disable it as per the feature creator @evanlin96069 's comment.

image


lipsanen:
tyvm

PR #301, opened 2023-05-04
Drawing entities and world

### LINKED ISSUE #268: GetPlayerVelocity returns nothing at demo playback

UncraftedName:
Fixes:
- Should fix #268. The hl2 client/server classes don't follow the expected naming convention, there may be other datamaps where this happens...
- Fixed `GetEnvironmentPortal()` returning the player entity in hl2
- I made a woopsie in the `DECL_FASTCALL` macro, luckily no one used it
- Someone (likely me) did a woopsie and hl2 hasn't had `spt_hud_ent_info` for a bit

Improvements:
- As always, more mesh rendering optimizations
- Made draw seams fully use the ray tracing from SDK
- Changed the tracing feature to use the DECL macros
- Portal placement doesn't draw if the trace didn't hit anything
- Added `utils::CachedField`. This is just a wrapper of `spt_entprops.GetFieldOffset(...)` that can be more convenient in some cases, and it means the user doesn't have to do manual offset adding and casting.

Additions:
- `spt_draw_map_overlay` (by evanlin)
- `spt_draw_world_collides` (and the hidden cvar `spt_draw_world_collides_mask`)
- `spt_draw_ent_collides`

`spt_draw_map_overlay` can be used to find traveling portal wrong warp locations.

image

`spt_draw_world_collides` (formerly draw_seams_v2) draws only the part of the world that the player is looking at, making it useful for finding seamshots. Also draws static props and displacements. This works by hooking into some of the low level tracing functions, it required the use of 3 SDK files.

image
image
image

`spt_draw_ent_collides` draws (almost) all entity collides, and tries to "guess" if the player will collide with that object. Also can be used to find server/vphysics desyncs as shown [here](https://youtu.be/Oq30Bnw14zc).

image
image


lipsanen:
Looks good to me. Can you write the release notes for the added commands/cvars?

UncraftedName:
- `spt_draw_map_overlay`: Draws brushes from other maps, can be used for finding traveling portal wrong warp setups (@evanlin96069)
- `spt_draw_world_collides` & `spt_draw_world_collides_mask` (hidden): Draws the piece of the world that the player is looking at, can be used for finding seamshots (@UncraftedName)
- `spt_draw_ent_collides`: Draws (almost) all entity collides, can be used for finding some client/server/vphysics desyncs (@UncraftedName)

lipsanen:
tyvm

PR #302, opened 2023-05-05
Sort datamap fields by offset

UncraftedName:
Sorting by offset is very useful for fields that might not exist in the datamap so that you can find the closest field. I also changed `spt_datamap_walk` to print every field with an offset, the logic for this wasn't quite right before.

Before:

image

After:

image


lipsanen:
tyvm

PR #303, opened 2023-05-06
Added spt_unload and optional path arg to spt_reload

UncraftedName:


spt/features/updater.cpp
```diff
@@ -341,24 +348,45 @@ bool Updater::ReloadPlugin()
 	return false;
 }
 
+void Updater::ReloadFromPath(const std::string& newPath)
+{
+	if (!fs::exists(newPath))
```
lipsanen, line 359: You should use the overload with the error code here. The non-error code overload is dumb and sometimes returns false but can also throw an exception under some conditions I forget
See: https://en.cppreference.com/w/cpp/filesystem/exists

Same goes for all the other filesystem functions, use the error code versions whenever possible.

lipsanen (changes requested):

spt/features/updater.cpp
UncraftedName: I could find docs on what error codes I should check for, so I just added the param and some warning messages to all functions. Good enough maybe? Probably reasonable to squash all 3 of these commits.

lipsanen: the main way I use the error codes is 
```
if(error == std::error_code()) { 
\\ no error, operation succeeded
}
```

lipsanen: so yea what you did seems good to me

lipsanen:
tyvm

PR #305, opened 2023-05-14
Add spt_hud_group

evanlin96069:
testchmb_a_080009

## HUD Group
Users can add and modify HUD groups. Each group can have its own HUD elements, position, font, and text color.
Old-style HUD will still be there so everything can still work in the same way.

## New CVars
- `spt_hud`: When set to 1, displays SPT HUD (old-style HUD).
- `spt_hud_position`: Turns on the position hud.
- `spt_hud_angles`: Turns on the angles hud with more precision.

## New Commands
- `spt_hud_group`: Modifies parameters in a given HUD group.
  - HUD element:
    `spt_hud_group <group name|all> add <HUD element> [args]`
    `spt_hud_group <group name|all> remove <index>`
    `spt_hud_group <group name|all> edit <index> <HUD element> [args]`
  - Display:
    `spt_hud_group <group name|all> <show|hide|toggle>`
  - Attributes:
    `spt_hud_group <group name|all> <x|y|pos|font|textcolor|background> [value]`

- `spt_hud_group_add`: `spt_hud_group_add <group name>`
- `spt_hud_group_remove`: `spt_hud_group_remove <group name|all>`
  

## Fix
- Fix `spt_saveloads` (breaks when `SignonStateSignal` is added in bb6d0acd6dc24ad9f213164c5b8ee27a4a8773b2)

Other Changes:
- Add autocomplete function for `spt_ihud_modify`.
- Change HUD callback sort keys to match command names.
- All HUD callback function now requires a `std::string args` so the same HUD element added to the HUD group can have different arguments (For example, different ent_info string or different flags_filter).
- `spt_hud_flags_filter` now only accept 1 argument. (So `spt_hud_flags_filter onground ducking` will become `spt_hud_flags_filter "onground ducking"`)
- Disable `spt_ihud` in BMS (crash).


lipsanen:
tyvm

PR #308, opened 2023-06-07
Mesh overflow detection

UncraftedName:
Chapter 34 of mesh rendering.

Users can now detect when they've added too much crap to a mesh.


lipsanen:
tyvm, does this have any user-facing benefits or is to help developers?

PR #314, opened 2023-08-05
Add spt_noclip_noslowfly

evanlin96069:
`spt_noclip_noslowfly` - Fix noclip slowfly.

 Sets the m_surfaceFriction to 1.0 every tick when noclip is enabled.

I think this fix is too small to be a new feature itself. So I added it in `noclip_nofix.cpp` and rename the file to `noclip_fixes.cpp`.
Although `portal_nogroundsnap.cpp` is also a noclip fix, it's kind of annoying to add all those `#ifndef OE` and `DoesGameLookLikePortal()`. So I didn't combine it into `noclip_fixes.cpp`.


lipsanen:
tyvm

PR #317, opened 2023-08-24
Add spt_on_portalled_pause_for

evanlin96069:
- `spt_on_portalled_pause_for` - Whenever player portalled, pause for this many ticks.

Small fix:
- Fix spt_jhud_velocity previous velocity not calculated on tick.


lipsanen:
I think 2838 added some afterticks thingy right? Would that be better for implementing the x ticks thing?

lipsanen:
tyvm

PR #319, opened 2023-09-02
spt_con_notify

thisis2838:
- Added `spt_con_notify` which allows the viewing of console messages on the top of the screen without having to enable `developer` mode.


lipsanen:
tyvm

PR #323, opened 2023-11-02
Add entity oob detection

UncraftedName:
Adds the following:
- `spt_hud_oob_ents 0|1` var
- `spt_draw_oob_ents 0|1|2|3` var
- `spt_print_oob_ents` cmd

This is a proof of concept entity oob checker using the OBB center. I used this point because it is what determines when non-player entities get teleported; during normal™ gameplay this point should not go oob. A small box is shown for oob entities, and the list can be shown in the hud or printed to console.
image
image
image

For now, I will say this is a proof of concept since it has not been officially decided how to determine if an ent is oob. Also, there is a blacklist of class names that are ignored in the entity collection - this list is probably not complete and may be better as a whitelist. The hope is that it will get updated over time if this feature is used and people report bugs.


lipsanen:
tyvm

PR #329, opened 2024-03-01
Add spt_drawline and other small fixes

evanlin96069:
`spt_drawline` - Draws line between two 3D Points.
Like `drawline` but with a color option and not debug overlay.
We can now harness the power of [Portal 2 tool](https://krzyhau.github.io/SeamshotCalculator/) by replacing the generated commands.

Other fixes:
Update some CI versions
Fix off-by-one error in pp grid.
Make spt_draw_world_collides update during tas_pause.


lipsanen:
tyvm, sry for the delay

PR #330, opened 2024-03-09
spt_noclip_persist, DoImageSpaceMotionBlur fix, Entropy Zero 2 fix, along with miscellanea

thisis2838:
- Added `spt_noclip_persist`: makes the player indefinitely float at the speed and in the direction of movement before entering noclip. Potentially useful for voidclip routing.
- Rewrote `spt_saveload`'s description to be more detailed with examples and clearer explanations.
- Fixed a bug where `DoImageSpaceMotionBlur`'s `pgpGlobals` was enumerated in `LoadFeature` through an offset from `ORIG` pointers, which no longer point to the original functions but rather the trampoline ones.
- Fixed `GameMovement` interface for Black Mesa.
- Fixed false positive for `DrawInternal`'s pattern `7467727-hl2` matching in Entropy Zero 2 (first release build). This led to crashes particularly when alt-firing.

### DIFF
--- a/spt/features/visualizations/renderer/internal/mesh_renderer.cpp
+++ b/spt/features/visualizations/renderer/internal/mesh_renderer.cpp
@@ -83,7 +83,9 @@ namespace patterns
 	         "7467727-hl2",
 	         "55 8B EC 81 EC A0 00 00 00 83 3D ?? ?? ?? ?? 00");
 	PATTERNS(CSkyBoxView__DrawInternal,
-	         "5135-portal1",
+	         "EZ2-1.0.0+original",
+	         "55 8B EC 81 EC 28 01 00 00 A1 ?? ?? ?? ?? 33 C5 89 45 ?? 8B 45 ??",
+			 "5135-portal1",
 	         "83 EC 28 53 56 8B F1 8B 0D ?? ?? ?? ??",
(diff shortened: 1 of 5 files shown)


spt/features/visualizations/renderer/internal/mesh_renderer.cpp
```diff
@@ -83,7 +83,9 @@ namespace patterns
 	         "7467727-hl2",
 	         "55 8B EC 81 EC A0 00 00 00 83 3D ?? ?? ?? ?? 00");
 	PATTERNS(CSkyBoxView__DrawInternal,
-	         "5135-portal1",
+	         "EZ2-1.0.0+original",
+	         "55 8B EC 81 EC 28 01 00 00 A1 ?? ?? ?? ?? 33 C5 89 45 ?? 8B 45 ??",
+			 "5135-portal1",
```
lipsanen, line 88: what's going on with this indenting over here

thisis2838: My IDE is constantly trying to make up its mind over whether to do whitespaces or tabs, and the auto-format also produces some wacky results

lipsanen:
tyvm and sry for the delay

PR #345, opened 2024-09-16
More stuff for the ImGui text HUD tab

UncraftedName:
I added all of the `spt_hud` cvars to the HUD tab in ImGui except for ent_info. It looks like this now:
Screenshot_24

Internally, the ImGui tab sorts cvars by their wrangled (legacy) name and `spt_hud` sorts by a custom sort key. Is it okay if I change the `spt_hud` logic to sort by name instead of using a sort key so that way they match up? We haven't been grouping the HUD cvars in any meaningful way so I think it should be fine to change the order.

### DIFF
--- a/spt/features/visualizations/imgui/spt_imgui_widgets.cpp
+++ b/spt/features/visualizations/imgui/spt_imgui_widgets.cpp
@@ -76,6 +76,22 @@ void SptImGui::HelpMarker(const char* fmt, ...)
 	va_end(va);
 }
 
+bool SptImGui::BeginBordered(const ImVec2& outer_size, float inner_width)
+{
+	if (ImGui::BeginTable("##table_border", 1, ImGuiTableFlags_BordersOuter, outer_size, inner_width))
+	{
+		ImGui::TableNextRow();
+		ImGui::TableNextColumn();
+		return true;
+	}
+	return false;
+}
+
+void SptImGui::EndBordered()
+{
+	ImGui::EndTable();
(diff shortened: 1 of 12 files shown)


lipsanen:
Any type of alphabetical sorting should be fine, I guess I'd be most concerned how to make it as easy as possible to find stuff from the user perspective.

spt/features/visualizations/imgui/spt_imgui_widgets.cpp
```diff
@@ -61,6 +62,23 @@ int SptImGui::CvarCombo(ConVar& c, const char* label, const char* const* opts, s
 	return val;
 }
 
+bool SptImGui::InputTextInteger(const char* label, const char* hint, long long& val, int radix)
+{
+	Assert(radix == 10 || radix == 16);
+	char buf[24];
+	const char* fmtSpecifier = radix == 10 ? "%" PRId64 : "%" PRIx64;
+	snprintf(buf, sizeof buf, fmtSpecifier, val);
+	buf[sizeof(buf) - 1] = '\0';
```
lipsanen, line 71: p sure this is redundant and should already be done by snprintf

UncraftedName: You're right, I can never remember which printf methods add the terminator and which don't so I just toss that in everywhere. Lemme fix.

lipsanen:
tyvm

PR #346, opened 2024-09-17
ImGui ent props

UncraftedName:
This is meant to be merged after #345.

Changes:
- Added `spt_hud_ent_info` to the hud tab
- Change `utils::ClientEntity()` to return null for large indices because the game doesn't do that check for us and this can trigger a crash otherwise (the game does check for `index<0` though)
- Small tweaks to `utils::FillInfoArray()` to make it slightly more resilient to weird user input

I've tested this with a bunch of weird user input - to the best of my knowledge it syncs up with the `utils::FillInfoArray()` logic and gracefully handles weird edge cases. When the user interacts with the GUI, it sort of "normalizes" the cvar value to something slightly more sensible.

https://github.com/user-attachments/assets/34f7f2ad-2303-4d97-854b-0eb20ec22341

### DIFF
--- a/spt/utils/ent_utils.cpp
+++ b/spt/utils/ent_utils.cpp
@@ -36,6 +36,8 @@ namespace utils
 {
 	IClientEntity* GetClientEntity(int index)
 	{
+		if (index >= interfaces::entList->GetHighestEntityIndex())
+			return nullptr;
 		return interfaces::entList->GetClientEntity(index + 1);
 	}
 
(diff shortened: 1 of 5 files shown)


spt/utils/ent_utils.cpp
```diff
@@ -318,31 +320,8 @@ namespace utils
...
-				++i;
-			}
-			else // error occurred
-			{
-				i = argSize;
-				arr = (wchar_t*)L"error occurred in parsing";
-				entries = 1;
-				break;
-			}
-
```
lipsanen, line 345: Any particular reason why the error check (else branch in the old code) was removed?

lipsanen:
tyvm

PR #349, opened 2024-09-20
A morbillion ImGui features

UncraftedName:
GUI Changes:
- Added QoL & cheat tab
- Added IHUD/JHUD tabs in addition to text HUD tab
- Added misc section to drawing tab
- Updated spt_hud_velocity HUD cvar to be a drop-down instead of a checkbox
- The tab callbacks in the main window are now in a child window (independent scrolling region)

Internal changes:
- Added RTA time to timer feature (this is so that you can tell if the feature is actually on when using it in ImGui)
- Removed the 'open' bool param from the main window ImGui callbacks - it's really inconvenient if the user has to check for it all the time and doesn't use it 95% of the time

The distinction between the QoL & the Cheats tab is pretty muddy. Some of that may be organized differently or shifted around in the future ¯\\_(ツ)_/¯.

Independent scrolling region:
image

QoL tab:
image

Cheats tab:
image

### DIFF
--- a/spt/features/collision_group.cpp
+++ b/spt/features/collision_group.cpp
@@ -1,24 +1,26 @@
 #include "stdafx.hpp"
 #include "ent_utils.hpp"
 #include "..\feature.hpp"
+#include "visualizations\imgui\imgui_interface.hpp"
 
 namespace patterns
 {
 	PATTERNS(SetCollisionGroup,
-		"4044",
-		"8B 54 24 ?? 8B 81 ?? ?? ?? ?? 3B C2",
-		"5135",
-		"56 8B F1 8B 86 ?? ?? ?? ?? 3B 44 24 08 8D",
-		"7467727",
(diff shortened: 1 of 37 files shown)


spt/features/collision_group.cpp
```diff
@@ -57,7 +67,25 @@ CON_COMMAND_F(y_spt_set_collision_group, "Set player's collision group\nUsually:
...
+			    ImGui::BeginDisabled(!playerPtr);
+			    if (ImGui::Button("set"))
+				    spt_collisiongroup.ORIG_SetCollisionGroup(playerPtr, (int)val);
+			    ImGui::EndDisabled();
+			    if (!playerPtr)
+				    ImGui::SetItemTooltip("server player not available");
+			    ImGui::SameLine();
+			    SptImGui::CmdHelpMarkerWithName(y_spt_set_collision_group_command);
+		    });
+	}
```
lipsanen, line 88: would this be better with a dropdown with the names of the collision groups? more effort though

lipsanen:
other than that looks good to me

UncraftedName:
As you wish
image

(I'm trusting the SDK values here)

lipsanen:
tyvm

PR #352, opened 2024-09-22
Get rid of the SDK submodules and move the code into the repo

lipsanen:
Redo of #351


UncraftedName:
This clang-format version crap is very tragic - your version is different from mine so you have a lot of unnecessary format diffs. I'll look into setting up a specific clang-format/uncrustify version, or we should completely nuke the formatting.

spt/cvars.cpp
```diff
@@ -325,59 +198,140 @@ void Cvar_RegisterSPTCvars()
...
 }
 
+
 void Cvar_UnregisterSPTCvars()
 {
-	if (!interfaces::g_pCVar)
+	if (!g_pCVar)
 		return;
 
+	ConCommandBase::s_pConCommandBases = NULL;
```
SirWillian, line 314: If you care about mimicking what the engine does naturally, this should be moved to the end of of Cvar_RegisterSPTCvars. Atm we are kinda living with a somewhat bogus pointer in this variable through the lifetime of the plugin since it's never cleared

spt/spt-serverplugin.cpp
```diff
@@ -132,22 +132,16 @@ void DefaultFOVChangeCallback(ConVar* var, char const* pOldString)
...
 CSourcePauseTool g_SourcePauseTool;
-EXPOSE_SINGLE_INTERFACE_GLOBALVAR(CSourcePauseTool,
-                                  IServerPluginCallbacks,
-                                  INTERFACEVERSION_ISERVERPLUGINCALLBACKS,
-                                  g_SourcePauseTool);
-#endif
+
+extern "C" __declspec(dllexport) const void* CreateInterface(const char* pName, int* pReturnCode) {
+	if (pReturnCode) {
+		*pReturnCode = 2007;
```
evanlin96069, line 139: According to the sdk
> if pReturnCode is set, it will return one of the following values (IFACE_OK, IFACE_FAILED)

Also, should we check for the `INTERFACEVERSION_ISERVERPLUGINCALLBACKS`?

spt/cvars.cpp
```diff
@@ -325,59 +198,140 @@ void Cvar_RegisterSPTCvars()
...
 		if (inittedCmd == cmd_to_feature.end())
 		{
 			DevWarning("Command %s was unloaded, because it was not initialized!\n", cmdName);
 			ConCommandBase* todelete = cmd;
 			cmd = (ConCommandBase*)cmd->GetNext();
-			RemoveCommandFromList(pluginCommandListHead, todelete);
+			RemoveCommandFromList(todelete);
 			continue;
 		}
+		else
```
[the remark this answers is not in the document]
lipsanen: Not in OE. The plugin unload does not get rid of the binary (the fix that works in 3420 hasn't been ported over there). So the commands need to be put back in the plugin list in order to work next time the plugin is loaded.

spt/cvars.cpp
lipsanen: The reasoning for this is that in OE the plugin list contains (at least?) one concommand after the cvars have been loaded normally. This caused a crash upon reloading the plugin

spt/spt-serverplugin.cpp
lipsanen: Checking for the version is not necessary. I also just set the return code to something funny since it isn't actually used.

spt/cvars.cpp
```diff
@@ -23,31 +23,21 @@ struct FeatureCommand
 	bool unhideOnUnregister;
 };
 static std::vector<FeatureCommand> cmd_to_feature;
-static bool first_time_init = true;
 static ConCommandBase** pluginCommandListHead = nullptr;
```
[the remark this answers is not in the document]
lipsanen: I think this was already removed but should be now at least

spt/features/autocomplete.cpp
```diff
@@ -39,17 +39,17 @@ bool AutocompleteFeature::ShouldLoadFeature()
...
 
 void AutocompleteFeature::InitHooks() {}
 
 void AutocompleteFeature::LoadFeature()
 {
 #ifndef OE
-	playdemoCommand = reinterpret_cast<ConCommand_guts*>(interfaces::g_pCVar->FindCommand("playdemo"));
-	loadCommand = reinterpret_cast<ConCommand_guts*>(interfaces::g_pCVar->FindCommand("load"));
-	execCommand = reinterpret_cast<ConCommand_guts*>(interfaces::g_pCVar->FindCommand("exec"));
+	playdemoCommand = reinterpret_cast<ConCommand_guts*>(g_pCVar->FindCommand("playdemo"));
```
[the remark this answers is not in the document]
lipsanen: they aren't really needed but I was too lazy to clean up everything

PR #364, opened 2024-11-25
Add Strafe HUD

evanlin96069:
Plot the strafe acceleration graph.
No imgui, just draw in vgui for now.

Also, moved DecodeUserCmdFromBuffer to playerio and made it a signal instead.
Changed `CreateMove` and `GetButtonBits` to use `DECL_HOOK_THISCALL` style declaration.

### DIFF
--- a/spt/features/strafehud.cpp
+++ b/spt/features/strafehud.cpp
@@ -1,0 +1,561 @@
+#include "stdafx.hpp"
+#include "..\feature.hpp"
+#include "hud.hpp"
+
+#ifdef SPT_HUD_ENABLED
+
+#include <immintrin.h>
+
+#include "interfaces.hpp"
+#include "playerio.hpp"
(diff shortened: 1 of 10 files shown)


spt/features/strafehud.cpp
```diff
@@ -0,0 +1,533 @@
...
+			accel /= -smallestAccel;
+		}
+	}
+
+	shouldUpdateLines = true;
+}
+
+void StrafeHUD::DrawHUD()
+{
+	interfaces::surface;
```
lipsanen, line 350: this looks like a leftover from something

lipsanen (changes requested):

spt/features/strafehud.cpp
evanlin96069: Fixed :)
Also added strafehud options in spt_gui

PR #365, opened 2024-11-25
Add spt_cam_draw_player

evanlin96069:
An option to not draw player model when camera control is enabled.
Requested by Eold. Sometimes you just want to use cinematic mode but don't want to see the player.

Also, removed `sv_cheats` check for camera control to be more consistent with other SPT features.


lipsanen:
tyvm

PR #371, opened 2025-03-12
Removed duplicate logic of spt_print_portals & spt_find_portals, changed to only spt_print_portals

UncraftedName:


lipsanen:
tyvm

PR #381, opened 2025-04-28
Player trace

UncraftedName:
Added commands:
- `spt_trace_start`/`spt_trace_stop`
- `spt_trace_next_tick`/`spt_trace_prev_tick`/`spt_trace_set_tick`
- `spt_trace_export`/`spt_trace_import`

Added cvars:
- `spt_draw_trace`
- `spt_hud_trace`
- `spt_trace_autoplay`
- `spt_trace_ent_collect_radius`
- `spt_trace_draw_portal_collision_entities`
- `spt_trace_draw_path_cones`
- `spt_trace_draw_cam_style`
- `spt_trace_draw_contact_points`
- Added `spt_cam_control 4` for viewing from the trace perspective

Planned changed for future:
- support for multiple active traces
- ImGui support for player trace

I know that source code is preferred over libraries, but the lzma SDK was too much of a pain to get working in SPT, so I compiled the project from source and added instructions (`thirdparty/xz/readme.txt`) on how others can compile them.


spt/features/camera.cpp
```diff
@@ -890,6 +894,14 @@ void Camera::HandleCinematicMode(bool active)
 	currentCam = InterpPath(currentTime);
 }
 
+void Camera::HandleTraceMode(bool active)
+{
+	if (!active)
+		return;
+	extern bool SptGetActiveTracePos(Vector & pos, QAngle & ang, float& fov);
```
lipsanen, line 901: Include the header preferably rather than declaring the function here separately

spt/features/visualizations/player_trace/tr_config.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+	enum TrPlayerCameraDrawType
+	{
+		TR_PCDT_FRUSTUM,
+		TR_PCDT_BOX_AND_LINE,
+
+		TR_PCDT_COUNT,
+	};
+
+	struct
```
lipsanen, line 33: Any particular reason why this file uses C-style structs?

lipsanen (changes requested):
Some parts of the code that stored entity pointers looked a bit spooky, hopefully it's made sure that pointers remain valid after/level loads/entitites getting destroyed etc.

spt/features/visualizations/player_trace/tr_config.hpp
UncraftedName: I'm not sure I understand the question.
- everything in that file is effectively a global, so it's all gotta be public (so I used structs instead of classes)
- the reason I used globals at all instead of hardcoding all the values is to smooth the making of the imgui interface for the player trace in the future (this way I can just edit those values directly from some separate file and it'll immediately effect how the player trace is drawn)
- the reason it's all in structs instead of being naked globals is because intellisense will make it a bit easier to use something like `trStyles.entities.drawObbCenter` or `trStyles.pathStyle.cones.tickInterval` than naked globals

UncraftedName:
> Some parts of the code that stored entity pointers looked a bit spooky, hopefully it's made sure that pointers remain valid after/level loads/entitites getting destroyed etc.

I'm not sure exactly which part of the code you're talking about, but if you mean the pointers in `TrEnt_v1` & `TrPhysicsObjectInfo_v1` - those are actually not *meant* to be valid, they are meant to serve as an ID for cache invalidation. For `IServerEntity*`, we have handles which *should* guarantee uniqueness (unless you reload thousands of times), so the pointer there is redundant.

There is no handle mechanism for `IPhysicsObject*` however, so I create my own ID to determine if the particular physics object is a a *new* one or one that I have seen before. One of the things that I use for that ID is the pointer, which of course is not guaranteed to be unique for every object (and in fact there's a special case for portals), but the hope is that it will maybe help.

spt/features/visualizations/player_trace/tr_feature.cpp
```diff
@@ -0,0 +1,343 @@
...
+	InitCommand(spt_trace_set_tick);
+	InitCommand(spt_trace_export);
+	InitCommand(spt_trace_import);
+
+	if (AddHudCallback("trace", [](auto) { spt_player_trace_feat.OnHudCallback(); }, spt_hud_trace))
+		SptImGui::RegisterHudCvarCheckbox(spt_hud_trace);
+
+	InitConcommandBase(spt_trace_autoplay);
+	InitConcommandBase(spt_trace_ent_collect_radius);
+	InitConcommandBase(spt_trace_draw_portal_collision_entities);
```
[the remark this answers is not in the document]
UncraftedName: Good catch, but only the last one. Fixed

spt/features/camera.cpp
UncraftedName: Alright, I declared all cvars in tr_config

spt/features/visualizations/player_trace/tr_config.hpp
lipsanen: I mean the name being at the end of the struct and not at the beginning?

UncraftedName: Is that really C-style structs? Those are just named instances of anonymous structs. Googles ai overview gives a good example:
```#include <stdio.h>

struct Employee {
    char name[50];
    struct { // Anonymous struct for address
        char street[50];
        char city[50];
        char state[3];
        int zip_code;
    } address;
};

int main() {
    struct Employee emp = {
        "John Doe",
        {
            "123 Main St",
            "Anytown",
            "CA",
            91234
        }
    };

    printf("Name: %s\n", emp.name);
    printf("Address: %s, %s, %s %d\n", 
           emp.address.street, emp.address.city, emp.address.state, emp.address.zip_code);

    return 0;
}
```

lipsanen:
tyvm

PR #384, opened 2025-05-05
Fixed SG overlay being broken for fizzled portals

UncraftedName:
I removed some logic during my ent list crap and didn't add it back cuz I'm stupid


lipsanen:
tyty

PR #385, opened 2025-05-12
Better trace versioning

UncraftedName:
The trace import/export logic was designed to be backwards compatible from the start, but I didn't really figure out the details of how I would do that until I tried to actually upgrade one of the 'lumps'. I made two main changes to the import logic:
- All of the main trace logic except the import stuff uses e.g. an alias `TrMap` instead of `TrMap_v1`. This is so that I don't have to rename huge chunks of trace code every time I upgrade a struct.
- There is now actual logic in place to upgrade individual 'lumps' in the trace via upgrade handlers.

I also went ahead and created a v2 for the player data so that it now stores vphysics velocity cuz I'm thinking of using that for a future project.


UncraftedName:
Don't merge yet, seems to have mega issues.

UncraftedName:
Fixed now, I was accidentally treating a large chunk of the cache as invalid causing the recording to be super slow and large.

UncraftedName:
Woops fixed another critical blunder where I didn't record the tick in the player data. Last fix probably :)

lipsanen:
I'm gonna lgtm this after a day if there's no other changes

PR #386, opened 2025-05-12
Attempt to fix multithreaded weirdness in ImGui callbacks

UncraftedName:
The ImGui window callbacks used to be called from `CShaderDeviceDx8::Present` which seems to have caused a bunch of undesirable multi-threaded complicated-ness. For example, we've seen that a debug build would sometimes result in some ImGui assert, but upon attaching a debugger the assert condition was true. The only way to fix that issue would be to have each feature that uses ImGui to own a mutex to prevent race conditions with most other feature callbacks (e.g. TickSignal). I think this is the *unfun* option.

Instead, I've separated the ImGui update and render logic.
- the ImGui update logic and window callbacks have been moved to FrameSignal
- `CShaderDeviceDx8::Present` calls `ImGui::Render()` and `ImGui_ImplDX9_RenderDrawData(ImGui::GetDrawData())`

This way, the Present hook will redraw the same ImGui data if FrameSignal isn't called. This comes with a couple caveats:
- during e.g. loads (when FrameSignal isn't called) it's not possible to update any ImGui windows
- also, after maximizing the game I can't show any ImGui data until FrameSignal is called again

These tradeoffs are well worth it imo to not have to deal with all ImGui features having their own mutex.

### DIFF
--- a/spt/features/visualizations/imgui/imgui_interface.cpp
+++ b/spt/features/visualizations/imgui/imgui_interface.cpp
@@ -3,12 +3,13 @@
 #include <set>
 
 #include "..\feature.hpp"
-#include "thirdparty/imgui/imgui.h"
 #include "interfaces.hpp"
 #include "spt/spt-serverplugin.hpp"
 #include "spt/utils/spt_vprof.hpp"
+#include "spt/utils/signals.hpp"
 #include "spt/features/hud.hpp"
 
+#include "thirdparty/imgui/imgui.h"
 #include "thirdparty/imgui/imgui_internal.h"
 #include "thirdparty/imgui/imgui_impl_dx9.h"
 #include "thirdparty/imgui/imgui_impl_win32.h"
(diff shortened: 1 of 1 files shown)


UncraftedName:
(fixed small typo in comment)

spt/features/visualizations/imgui/imgui_interface.cpp
```diff
@@ -95,9 +99,13 @@ class SptImGuiFeature : public FeatureWrapper<SptImGuiFeature>
...
 		InitDx9,
 		OverrideWndProc,
+		FrameSignalCalledAtLeastOnce,
 
 		Loaded = OverrideWndProc,
-	} inline static loadState = LoadState::None;
+	};
+
+	inline static std::atomic<LoadState> loadState = LoadState::None;
+	inline static std::unique_ptr<std::recursive_mutex> imguiMutex = nullptr;
```
lipsanen, line 108: Is the unique_ptr required here? Wouldn't a normal std::recursive_mutex work as well?

UncraftedName: You're right. If it's not static it would be required to be moveable since features must be moveable. But since it's static I don't think I need the unique_ptr there.

lipsanen:
So previously the issue was that the "main thread" could be running at the same time as the ImGui code from CShaderDeviceDx8::Present?

UncraftedName:
Yeah that's exactly right

> So previously the issue was that the "main thread" could be running at the same time as the ImGui code from CShaderDeviceDx8::Present?

Yeah as far as I can tell that's the main cause of the weirdness.

lipsanen:
ok gotcha, ty

PR #394, opened 2025-06-24
Remove json lib from PCH

UncraftedName:
Decreased compile time of Debug ful rebuild on my laptop from 61s to 44s. Plus I found the source of a couple weird macro conflicts with the json lib and pulled it out of the PCH (always thought it was weird that it was in there).

For future pull requests I think it's a good idea to ask people to keep STL/system headers in stdafx.hpp.

### DIFF
--- a/SDK/bms_new/include/public/video/ivideoservices.h
+++ b/SDK/bms_new/include/public/video/ivideoservices.h
@@ -27,9 +27,7 @@
 #endif
 
 
-#ifndef nullptr
-	#define nullptr		( 0 )
-#endif
+
 #ifndef INT32_MAX
 #define INT32_MAX    (0x7FFFFFFF)
 #endif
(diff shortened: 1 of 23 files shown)


SirWillian:
Shouldn't we keep the headers in the cpp files for IWYU reasons? Include guards should prevent the builds slowing down (unless they don't)

UncraftedName:
After some discussion with bill, I ended up not removing the STL headers from feature files. I just removed some weird incompatibility with the json lib and the SDK. In the future I may try to go through and make each file IWYU and make the PCH only an optimization, but there's some weird errors I keep running into with the <algorithm> header.

From what I've messed around with, removing the PCH entirely does significantly slows down compile times, but I didn't notice a significant improvement when I added in a bunch of STL headers, so I'm keeping it *mostly* as is for now.

PR #398, opened 2025-07-11
Signal for portal teleports

UncraftedName:


UncraftedName:
I realized the naked hooks macros are not actually used in this PR (I made them without realizing that), but they still might be used in some future feature.

UncraftedName:
added the VA_OPT variant

PR #399, opened 2025-07-15
Decouple player trace & serialization util code

UncraftedName:
This PR only has QoL dev features.

While working on FCPS serialization I needed to reuse some of the serialization logic in the player trace code. I've decoupled the logic and moved the serialization classes to `utils/serialize.hpp`. I've created a fully functional serialization system with functionality for:
- `std::istream`/`std::ostream` wrappers
- reading/writing to/from memory
- reading/writing using lzma compression

I've also removed the footer that I appended to xz streams for the player trace. I may still try to use third party serialization libraries for future features - the main purpose of this PR is to decouple the serialization & player trace logic.


UncraftedName:
rebased against master

spt/features/visualizations/player_trace/import_export/tr_binary.hpp
```diff
@@ -4,54 +4,159 @@
...
+		// a bunch of info from when the trace was first recorded
+
+		struct
+		{
+			char gameName[16]; // max length returned by utils::GetGameName()
+			char modName[260]; // MAX_OSPATH
+			int32_t gameVersion;
+		} gameInfo;
+
+		char playerName[32]; // MAX_PLAYER_NAME_LENGTH
```
lipsanen, line 52: I think these char arrays are missing one byte to account for the null terminator. strncpy does not null-terminate the string if it runs out of space. Very comical design decision on their part, got em good.

lipsanen (changes requested):

PR #401, opened 2025-07-25
Simplify shadow feature

UncraftedName:
The old version relied on hooks but the player's physics object is accessible through the SDK (plus a datamap). Since the getters/setters are really simple I've also inlined usages of the shadow feature.


lipsanen (approved):

lipsanen:
tyvm

PR #403, opened 2025-07-29
FCPS

UncraftedName:
# User facing changes:
- added `spt_fcps_record` to enable recording of FCPS events
- added `spt_fcps_max_queue_size` to set how many FCPS events can be stored in memory
- added `spt_fcps_dump_events` to see recorded & loaded FCPS events
- added `spt_draw_fcps_events`
- added `spt_fcps_export`/`spt_fcps_import`
- added `spt_fcps_next_step`/`spt_fcps_prev_step`
- added `spt_fcps_draw_detail` & `spt_fcps_animation_hz`
- the swept box visualization has been changed slightly to minimize z-fighting with itself
- the player trace now records ticks when FCPS happens

# Internals

## The FCPS queue
When FCPS recording is enabled, each time FCPS happens SPT will store everything that happens in an `FcpsEvent`. To limit the memory usage of this feature, events are added to an `std::deque`. Whenever events are recorded or imported, old events are evicted if the size of the queue is above `spt_fcps_max_queue_size`. Bill and I discussed the specifics of this, and ultimately we landed on two exceptions:
1) imported events bypass the max queue size (until more events are recorded or imported)
2) events that are currently being animated are saved until animation is done

If the size of the queue is bigger than the limit after importing events, `spt_fcps_dump_events` shows which events will be evicted next.

## Serialization
After banging my head on the wall with the player trace serialization, I realized I wanted some standardized format so that I could parse the files outside of SPT without rebuilding a parser. For this I choose [flatbuffers](https://github.com/google/flatbuffers):
- it's schema based (like protobuf)
- it's fast, all of the data is stored inline
- the library itself and the auto-generated code is a pretty reasonable size

It is a little bit awkward to build messages, but of the 4 options I considered (the others being json, protobuf, & cap'n'proto), it seems to be the best fit for my needs. I will use it for other features in the future.

### Compact meshes
One of the things I wanted to store was the mesh of entities that were hit with a trace. For this I created a class called MbCompactMesh - this class stores duplicate vertices to minimize mesh size. You can build it from the ground up or you can have the mesh builder populate an instance for you after using the mesh builder primitives. I incorporated this into the mesh builder for static meshes and added a new mesh renderer test case.

# Images

A 'non-detail view' of a single iteration of an FCPS event
<p align="center">
image
</p>

A 'detail view' of the same event (set with `spt_fcps_draw_detail 1`, allowing you to scrub through the ray traces for each corner pair)
<p align="center">
image
</p>

The last (successful) entity trace from the same event (yellow - trace start, green - trace hit, turquoise - trace target)
<p align="center">
image
</p>

Planetary alignment (my beloved)
<p align="center">
image
</p>

FCPS on a turret
<p align="center">
image
</p>

Example of the FCPS queue
<p align="center">
image
</p>

The mesh test for MbCompactMesh
<p align="center">
image
</p>


lipsanen:
tyvm

PR #406, opened 2025-08-07
Add sound and physics hook timer rng fixes

evanlin96069:
From https://github.com/UncraftedName/SourcePauseTool/tree/rng-logging


lipsanen:
tyvm

PR #421, opened 2025-11-07
Add env var override build number detection

evanlin96069:
Add `SPT_USE_BUILD_NUMBER` for overriding build number detection.
Mainly for leaked builds or temporary workaround when the build number detection isn't correct.


UncraftedName:
woops I thought this was an issue so I added the labels xD

PR #424, opened 2025-11-07
Fix spt_pause 2 crashing in 2707

### LINKED ISSUE #418: spt_pause 2 crashed in hl2 2707
Reported by peng
Works fine in 4044, crashes in 2707.

tmob03:
I don't know why the original comment said hls 2707 is 222, it isn't and it was crashing there too. So now it works in both hl2 and hls 2707.
Closes #418


evanlin96069 (changes requested):

lipsanen:
tyty

PR #428, opened 2025-11-14
Convert project to use CMake

lipsanen:
Could still mess with the CI a little bit if we want, since I made the CI only have Debug/Release rather than have every combination of Debug OE, Release OE, Debug (2007), Release (2007), etc. This makes it a bit simpler but also increases the CI time from 3~ minutes to 12~ minutes, since it compiles all 4 engine branches in one config.


README.md
```diff
@@ -37,23 +37,16 @@ A plugin for all your pausing needs.
 1. Open Visual Studio 2022. Click on Tools → Get Tools and Features... from the top bar of the window.
 <br>This should open the Visual Studio Installer in another window. From the Workload tab, install `Desktop development with C++`. From the Individual Components tab, install:
     * MSVC v143 - VS 2022 C++ x64/x86 build tools
+	* C++ CMake tools for Windows
 2. Run the following in cmd:
     ```
     git clone --recurse-submodules https://github.com/OutOfBoundsOffice/SourcePauseTool.git
     SourcePauseTool\spt.sln
```
evanlin96069: No more spt.sln

PR #433, opened 2025-11-21
Fix ground detection

evanlin96069:
Fixed incorrect ground detection bug when `toggle_duck` on complex brushes or some static props.

Code matches `CGameMovement::CanUnduck` using `trace.startsolid`

Bumped `tas_strafe_version` to 10.


lipsanen:
tyty

PR #435, opened 2025-11-24
Fixes #434

### LINKED ISSUE #434: spt_hud_ent_info 0,.* crashes on Portal 3420
Other ent indices above 0 seem to work fine, regardless of whether there's an entity on that index at that time. On 5135, it displays ent info for CWorld. Haven't tested other games or versions

UncraftedName:
UntitledParser uses different send prop types for protocol versions 14 & below than for 15 & above. As far as I remember this corresponds with version 5135 (which is protocol 15) - that has a DPT_VectorXY field. This adjustment takes care of strings which previously showed up as "enable to parse" since they were treated as `DPT_Array`. Also, the cast for DPT_String was incorrect which fixes #434.

Tested `spt_hud_ent_info 0,.*` on portal 5135 & 3420.

### DIFF
--- a/SDK/orangebox_new/include/public/dt_common.h
+++ b/SDK/orangebox_new/include/public/dt_common.h
@@ -103,6 +103,7 @@ typedef enum
 	DPT_Int=0,
 	DPT_Float,
 	DPT_Vector,
+	DPT_VectorXY, // Only encodes the XY of a vector, ignores Z
 	DPT_String,
 	DPT_Array,	// An array of the base types (can't be of datatables).
 	DPT_DataTable,
--- a/spt/features/ent_props.cpp
+++ b/spt/features/ent_props.cpp
@@ -28,6 +28,11 @@ ConVar y_spt_hud_ent_info(
     "Display entity info on HUD. Format is \"[ent index],[prop regex],[prop regex],...,[prop regex];[ent index],...,[prop regex]\".\n");
 #endif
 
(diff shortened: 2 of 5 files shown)


UncraftedName:
Seems to be broken on 4104
image

UncraftedName:
I finally dug into this a bit more.

In addition to the fix mentioned above, many of the mysteries of datatables are no longer as mysterious if you see that some of the properties have proxy functions. These proxy functions may:
- grab only some bytes of the field (e.g. some props which are `DPT_Int` are ints, some are bools (`RecvProxy_Int32ToInt8`), etc)
- adjust the data (e.g. `m_flSimulationTime` goes through `RecvProxy_SimulationTime`)
- call functions on the entity and write directly to the entity field

Because of the last point, it's not possible to call these proxy functions to obtain the "actual" values, so some of them appear broken when displayed. There's also a lot of them, so it's not viable to write custom wrappers ([like I tried to do in UntitledParser](https://github.com/UncraftedName/UntitledParser/blob/master/DemoParser/src/Parser/EntityStuff/EntityPropertyEnumManager.cs)).

Anyways, I also added the ability to iterate through recursive datatables but disabled it by default because there's a lot of garbage properties. This can be enabled with `spt_hud_ent_info_recurse 1`.

With `spt_hud_ent_info_recurse 0`:
image


With `spt_hud_ent_info_recurse 1`:
image

The player's fields with recursion:
image

Tested on 3420, 4104, 5135, & steampipe.

PR #450, opened 2026-03-14
Monocle support

UncraftedName:
# Summary

[Monocle](https://github.com/UncraftedName/Monocle) is a library I developed to figure out where VAGs are possible. You tell it the pos/ang of your portals, the player position, if the map origin is inbounds, the game version, how many ants there are in your neighbors house, a few other things, and it tells you if a VAG is possible. If you create an image and repeat this for a bunch of different points on a portal, then color each pixel where you entered the portal some color based on the result, you can create some crazy looking images.

https://github.com/user-attachments/assets/eac37e75-700e-43e0-aab9-2818372f8186

The tool allows you to click on a particular pixel and it'll give you a setpos command for that point.

It turns out it's possible to do more exotic angle glitches (in layman's terms you might call these "double VAGs" or "negative VAGs"). This tool makes it possible to find them.

Here's one in 09:
image
image

Here's one found by xeonic:
image
image

# Changes
- Add new VAG testing tool in `spt_gui` under `Drawing/VAG test`
- Removed `spt_vag_search`


UncraftedName:
forgot to add it to the imgui window and accidentally left an old file in

UncraftedName:
some suggestions by bill & evanlin

PR #456, opened 2026-03-24
Add null checks for cinput_thisptr

evanlin96069:
Fix strafehud crash


evanlin96069:
Initialized the return value.
Removed the version 9 check in `GetPlayerData` because it's unnecessary. All members of `data` got initialized right after.
Fixed `data.Basevelocity` got initialized with `Vector()`. `Basevelocity` seems only used in `Strafe::Move`, which is only used in `SimulateFrames`, which is used by no one, so this shouldn't affect anything.

PR #458, opened 2026-04-05
Monocle - effective teleport space

UncraftedName:
Only some spaces in the portal hole are usually possible to VAG on
<p align="center">
image
</p>


lipsanen:
tyvm

PR #465, opened 2026-04-25
Add spt_disable_portal_roll_offset

evanlin96069:
For the new Portal rule
> Whether the player is oob will be determined based on the player's camera position disregarding drunk cam

No Steampipe support yet because the function in question got inlined there. We don't often have this problem (VAG clipping glitch) when verifying Steampipe runs anyway.

Also, fixed the GetPlayerMins/Maxs signature. All x86 ABIs require that for returning structs by value we copy the sret argument into %rax/%eax (depending on ABI) for the return.

### DIFF
--- a/spt/features/game_fixes/visual_fixes.cpp
+++ b/spt/features/game_fixes/visual_fixes.cpp
@@ -3,6 +3,7 @@
 #include "..\feature.hpp"
 #include "convar.hpp"
 #include "ent_utils.hpp"
+#include "game_detection.hpp"
 #include "signals.hpp"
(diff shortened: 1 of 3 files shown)


lipsanen:
I think an alternative way that would not require new hooks would be to modify the CViewSetup in the render function that the overlay hooks and you could check for this cvar in the oob detector code. idk which is better here

lipsanen:
Ah I see you were talking about alternative approaches in the discord with bill, I guess this is good to go then

lipsanen:
tyty

PR #470, opened 2026-06-19
Fix UTIL_TraceRay signature in 2013

evanlin96069:
`spt_hud_oob 1` was causing some random crashes in Portal steampipe because of this.


lipsanen:
tyvm

PR #474, opened 2026-07-29
Added spt_set_game_seed_on_load

Ramalayha:
Similiar to ivp seed but for game rng such as the spin of thrown grenades in hl2.


UncraftedName:
If we are going through with #473, then I would remove the FCVAR_CHEAT flag

evanlin96069:
Also add your cvar in the template script. https://github.com/OutOfBoundsOffice/SourcePauseTool/blob/3dfdd68314a9a47e2dc969706234814eacbf9cc7/spt/features/tas.cpp#L422

SDK/orangebox_new/lib/vstdlib.lib
evanlin96069, line 1: Do we still need the .lib file when we're using GetProcAddress to get the function?

Ramalayha: Oops, no we shouldnt i just forgot to delete it

PR #475, opened 2026-07-30
Fixed spt_set_game_seed_on_load

Ramalayha:
Applied the changes suggested in the previous pull request [#474](https://github.com/OutOfBoundsOffice/SourcePauseTool/pull/474)


spt/features/game_fixes/rng.cpp
```diff
@@ -30,6 +34,12 @@ ConVar spt_set_all_sounds_available_after_load(
     FCVAR_CHEAT | FCVAR_TAS_RESET,
     "Set to 1 for consistent sound rng, which may contribute to the uniform random stream. Useful for new TAS scripts, but may break old scripts.");
 
+ConVar spt_set_game_seed_on_load(
+    "spt_set_game_seed_on_load",
+    "",
+    FCVAR_TAS_RESET,
```
evanlin96069: This should probably be `FCVAR_NONE` to be consistent with other set on load cvars.

spt/features/game_fixes/rng.cpp
```diff
@@ -102,7 +112,10 @@ void RNGStuff::LoadFeature()
 	if (ORIG_CBasePlayer__InitVCollision)
 	{
 		if (ORIG_ivp_srand && spt_rng.IVP_RAND_SEED)
+		{
 			InitConcommandBase(y_spt_set_ivp_seed_on_load);
+			InitConcommandBase(spt_set_game_seed_on_load);
```
evanlin96069: This shouldn't be under the ivp seed condition. It should be under like `if (SetRandomSeed)` since SetRandomSeed can be null if the symbol somehow not found.

spt/features/game_fixes/rng.cpp
```diff
@@ -146,6 +159,17 @@ IMPL_HOOK_THISCALL(RNGStuff,
 			y_spt_set_ivp_seed_on_load.SetValue("");
 		}
 		DevWarning("spt: ivp seed is %u\n", *spt_rng.IVP_RAND_SEED);
+
+		if (spt_set_game_seed_on_load.GetString()[0] != '\0')
```
evanlin96069, line 522: Same here. Should be under the `if (SetRandomSeed)` check instead of `if (spt_rng.ORIG_ivp_srand && spt_rng.IVP_RAND_SEED)`. Tbh, I'm not sure if this is the correct timing to set RNG, but I guess if it works it should be fine.

evanlin96069 (changes requested):

